PR #4, opened 2020-07-02
Call exitHandler.Pass() when passing detection

kamaln7:
This calls `exitHandler.Pass()` when passing detection.

### DIFF
--- a/detect.go
+++ b/detect.go
@@ -166,4 +166,6 @@ func Detect(detector Detector, options ...Option) {
 			return
 		}
 	}
+
+	config.exitHandler.Pass()
 }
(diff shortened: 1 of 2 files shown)


nebhale:
@kamaln7 Thanks for the first-time contribution.  I was wondering, what prompted you to find this bug?  In all cases, this has no net effect (the detect process will exit 0), but is definitely worth fixing.

kamaln7:
Thank you, @nebhale! We use an exit handler in our buildpacks' detection tests that relies on calls to `Pass()` and `Fail()` to check if the detection results are correct.

nebhale:
Fix tagged and released as `1.14.0`.

PR #6, opened 2020-07-17
Don't trim spaces

ekcasey:
There are cases where spaces may be intentionally included in environment variables.

Signed-off-by: Emily Casey <ecasey@vmware.com>


ekcasey:
Updated to trim all whitespace from bindings but not from the environment.

PR #23, opened 2020-10-27
Remove BurntSushi

nebhale:
Previously, this project used the BurntSushi TOML library.  This library is now unmaintained and shouldn't be used for long term projects.  This change migrates the project onto an alternative TOML library.

### DIFF
--- a/build.go
+++ b/build.go
@@ -18,12 +18,13 @@ package libcnb
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"reflect"
 	"strings"
 
-	"github.com/BurntSushi/toml"
+	"github.com/pelletier/go-toml"
 
 	"github.com/buildpacks/libcnb/internal"
 	"github.com/buildpacks/libcnb/poet"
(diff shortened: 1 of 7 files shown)


ekcasey (approved):

nebhale:
Agreed to hold this for the next version of `go-toml` with fixes for known quoting issues.

samj1912:
v1.9.0 was just released. Should this be merged now?

ekcasey:
@sclevine was following the bug/issues with go-toml closely, maybe he can give us his take on whether it is ready for primetime?

samj1912:
The other thing to note is that we have different projects using different toml libraries. From what I can see lifecycle uses Burnt Sushi where as pack uses go toml. Similarly libpak (not in CNB but it is used alongside libcnb quite a bit) which consumes libcnb uses go toml. 

Maybe out of scope for this PR but we should also decide on a CNB org level which toml library should we use to avoid duplicate libraries being included. 

cc @buildpacks/platform-maintainers @buildpacks/implementation-maintainers for their thoughts on the usage of these two libraries.

ekcasey:
@samj1912 Since BurntSushi is no longer maintained I think the plan is to switch to go-toml. However, we needed to wait for some bugfixes in go-toml to avoid introducing bugs into our applications

samj1912:
Given that Burnt sushi toml was revived - should we close this?

samj1912:
Closing for now since we merged a bunch of dependabot changes to bump the burnt sushi toml version. Feel free to re-open if we feel strongly about moving to this.

Issue #48, opened 2021-04-07
Support Multiple Buildpack APIs


ekcasey:
A single release of `libcnb` should support multiple buildpack APIs so that consumers can upgrade to the latest to pull in bugfixes and improvements without simultaneously upgrading to the newest API.

samj1912:
#54 starts adding back support for 0.5 again.

samj1912:
Fixed in #54

PR #54, opened 2021-04-15
Modifications to enable libcnb to be compatible with both versions 0.5 and 0.6 of the API

dmikusa-pivotal:
At the moment, the latest version of libcnb (1.20.0) breaks compatibility for buildpacks that are still on the 0.5 version of the API. This patch creates compatibility for both 0.5 and 0.6.

The use case for this change is upgrading buildpacks. On the Paketo team, we have quite a few buildpacks that are using libpak and libcnb. This abrupt change means we either have to hold back libcnb updates or we have to go through and update all of the buildpacks to use the new API. Having compatibility for both version 0.5 and 0.6 eases the transition as we can keep updating libcnb and migrate buildpacks to 0.6 at a more leisurely pace.

My suggestion would be to remove this 0.5 compatibility layer when we add support for the 0.7 (or whatever's next) API version. That way we're supporting the latest two API versions at any given moment, again helping buildpack authors to transition more easily.

### DIFF
--- a/build_test.go
+++ b/build_test.go
@@ -17,12 +17,13 @@
 package libcnb_test
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"strings"
 	"testing"
+	"text/template"
 
 	. "github.com/onsi/gomega"
 	"github.com/sclevine/spec"
@@ -49,6 +50,7 @@ func testBuild(t *testing.T, context spec.G, it spec.S) {
 		layersPath        string
 		platformPath      string
 		tomlWriter        *mocks.TOMLWriter
+		buildpackTOML     *template.Template
 
 		workingDir string
 	)
@@ -68,7 +70,7 @@ func testBuild(t *testing.T, context spec.G, it spec.S) {
 		Expect(os.Setenv("CNB_BUILDPACK_DIR", buildpackPath)).To(Succeed())
 
 		bpTOMLContents = `
-api = "##API_VERSION##"
+api = "{{.APIVersion}}"
 
 [buildpack]
 id = "test-id"
@@ -99,10 +101,14 @@ mixins = ["test-name"]
 [metadata]
 test-key = "test-value"
 `
-		Expect(ioutil.WriteFile(filepath.Join(buildpackPath, "buildpack.toml"),
-			[]byte(strings.Replace(bpTOMLContents, "##API_VERSION##", "0.6", 1)),
-			0644),
-		).To(Succeed())
+		buildpackTOML, err = template.New("buildpack.toml").Parse(bpTOMLContents)
+		Expect(err).ToNot(HaveOccurred())
+
+		var b bytes.Buffer
+		err = buildpackTOML.Execute(&b, map[string]string{"APIVersion": "0.6"})
+		Expect(err).ToNot(HaveOccurred())
+
+		Expect(ioutil.WriteFile(filepath.Join(buildpackPath, "buildpack.toml"), b.Bytes(), 0644)).To(Succeed())
 
 		f, err := ioutil.TempFile("", "build-buildpackplan-path")
 		Expect(err).NotTo(HaveOccurred())
@@ -197,7 +203,7 @@ version = "1.1.1"
 			)
 
 			Expect(exitHandler.Calls[0].Arguments.Get(0)).To(MatchError(
-				"this version of libcnb is only compatible with buildpack API 0.5 and 0.6",
+				"this version of libcnb is only compatible with buildpack APIs 0.5 and 0.6",
 			))
 		})
 	})
@@ -394,10 +400,11 @@ version = "1.1.1"
 	})
 
 	it("writes 0.5 layer metadata", func() {
-		Expect(ioutil.WriteFile(filepath.Join(buildpackPath, "buildpack.toml"),
-			[]byte(strings.Replace(bpTOMLContents, "##API_VERSION##", "0.5", 1)),
-			0644),
-		).To(Succeed())
+		var b bytes.Buffer
+		err := buildpackTOML.Execute(&b, map[string]string{"APIVersion": "0.5"})
+		Expect(err).ToNot(HaveOccurred())
+
+		Expect(ioutil.WriteFile(filepath.Join(buildpackPath, "buildpack.toml"), b.Bytes(), 0644)).To(Succeed())
 
 		layer := libcnb.Layer{
 			Name: "test-name",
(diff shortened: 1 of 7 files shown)


layer.go
```diff
@@ -146,5 +147,17 @@ func (l *Layers) Layer(name string) (Layer, error) {
 		return Layer{}, fmt.Errorf("unable to decode layer metadata %s\n%w", f, err)
 	}
 
+	if !layer.Build && !layer.Cache && !layer.Launch {
+		// if all three are false, that could mean we have a API <= 0.5 TOML file
+		// try parsing the <= 0.5 API format where these were top level attributes
+		buf := internal.LayerAPI5{}
+		if _, err := toml.DecodeFile(f, &buf); err != nil && !os.IsNotExist(err) {
+			return Layer{}, fmt.Errorf("unable to decode layer metadata %s\n%w", f, err)
```
samj1912, line 155: Given that we are doing a best effort decode here, should we be propogating this error upwards or should we just ignore it?

internal/layer_api_5.go
```diff
@@ -0,0 +1,16 @@
...
+// LayerAPI5 is used for backwards compatibility with serialization/deserialization of the layer toml
+type LayerAPI5 struct {
+	// Build indicates that a layer should be used for builds.
+	Build bool `toml:"build"`
+
+	// Cache indicates that a layer should be cached.
+	Cache bool `toml:"cache"`
+
+	// Launch indicates that a layer should be used for launch.
+	Launch bool `toml:"launch"`
```
samj1912, line 12: It might be easier to keep this as LayerTypes and embed LayerTypes in LayerAPI5.
This was what was done previously in api v0.5. This should make the conversions between v0.5 and v0.6 easier?

See https://github.com/buildpacks/libcnb/blob/ab7d1d48e082a754434ce52c941ece92194e4149/layer.go#L73

samj1912 (commented):
There were some additional attributes added to launch.toml like the default flag for processes. Should we add a warning or throw an error if someone sets that with the v0.5 api?

internal/layer_api_5.go
dmikusa-pivotal: I tried that initially, but it creates a circular package reference, internal -> libcnb -> internal -> ... 

I broke that by just including the three bools directly. It's not DRY, but this data structure will go away with the next API version, so it doesn't seem too awful. I also wanted to put this struct into the internal package to keep someone from trying to use it directly. I could move it next to the Layer struct in layer.go and use LayerTypes embedded. It seems unlikely that with a doc comment, someone would use it directly. 

Let me know if you have a preference, I'm OK with either one.

layer.go
dmikusa-pivotal: When I was playing around with this, it seems like the 0.6 API layer metadata struct can be used with parsing either the layer 0.5 or 0.6 format TOML. It's doesn't read the correct values, but it's structurally fine and the TOML library won't throw an error.

What's broken is that if the API 0.6 struct is used to parse the layer 0.5 TOML (or vise versa), you'll miss the build, launch, cache properties because they are in a different place. It doesn't cause an error, I think because we're not using strict parsing, but those fields will all end up being the default value, hence my use of build=false, launch=false, cache=false as the trigger to try parsing using the 0.5 API struct.

I think we should propagate up errors for both parsing attempts because, as far as I can tell, a property formatted layer TOML file shouldn't cause either parsing attempt to fail and return an error. If there was an error, I think it would put libcnb into a state where we shouldn't continue. If we just logged the error and continued on, it could mean that build=false, launch=false, cache=false is set set for a layer where the metadata is supposed to indicate something else.

Anyway, that's my thought process. I'm open to suggestions if you think there's a situation where we could benefit from logging the error and continuing processing.

samj1912: That makes sense - the only thing that slightly bugs me is that fact that the layer flags are always set to false starting API 0.6 per https://buildpacks.io/docs/reference/spec/migration/buildpack-api-0.5-0.6/#opt-in-layer-caching-and-launchtoml-format-changes and this was done to avoid issues where leftover layers were being persisted even after they were removed from the buildpack code. In the spirit of this RFC does it make sense to avoid doing this decode entirely or should we still attempt to do it as it was part of the v0.5 api. The other thing to note is that the v0.6+ case will always hit this condition and attempt to decode it using the old api.

I guess one alternative (which may not be worth the effort) is to make `Layers`, BP API aware when we construct it in `BuildContext` and have it decode it using this internal struct only when the API version is 0.5. LMK what you think.

dmikusa-pivotal: >In the spirit of this RFC does it make sense to avoid doing this decode entirely or should we still attempt to do it as it was part of the v0.5 API.

I don't think this PR changes the behavior that we currently have with the 0.6 API support. If you have a 0.6 API buildpack and it sets launch=false, build=false, cache=false on a layer, that'll still be what get's picked. The only difference is that there will be a second attempt to reach the 0.5 API layer. In that case, if this is legitimately a 0.6 API layer with launch=false, build=false, cache=false will come back false because the API 0.6 layer TOML file won't have the properties in the right location to be parsed into the 0.5 API struct. We do assign them back to the struct, but they'll end up all being false so in the end launch=false, build=false and cache=false.

I put this together mostly as a way to get things working for now. Long term, if we are going to support multiple API layers, what we need is an easier way to access the API version, that way it can decode specifically for the API being used and not have to guess.

My preference would be to handle that as part of #48 and go with this for now. Would that be OK?

internal/layer_api_5.go
dmikusa-pivotal: Done

layer.go
samj1912: SGTM!

dmikusa-pivotal:
>There were some additional attributes added to launch.toml like the default flag for processes. Should we add a warning or throw an error if someone sets that with the v0.5 api?

samj1912 (approved):

ekcasey (approved):

build_test.go
```diff
@@ -64,9 +67,8 @@ func testBuild(t *testing.T, context spec.G, it spec.S) {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(os.Setenv("CNB_BUILDPACK_DIR", buildpackPath)).To(Succeed())
 
-		Expect(ioutil.WriteFile(filepath.Join(buildpackPath, "buildpack.toml"),
-			[]byte(`
-api = "0.6"
+		bpTOMLContents = `
+api = "##API_VERSION##"
```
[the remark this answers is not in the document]
dmikusa-pivotal: Done. See latest commit.

ekcasey (approved):

PR #57, opened 2021-05-13
Add support for k8s service bindings during build time

### LINKED ISSUE #55: Setting `SERVICE_BINDING_ROOT` does not at build time
When you run a container created by buildpacks, you may set an env variable `SERVICE_BINDING_ROOT` to instruct where to find bindings. This request is to support the same at build time.

Presently if you pass `--env SERVICE_BINDING_ROOT` to `pack build` (or similar tools) for libcnb based buildpacks, the value is ignored and it requires you to locate your bindings under `/platform/bindings` or they are ignored.

samj1912:
Signed-off-by: Sambhav Kothari <skothari44@bloomberg.net>

Fixes #55 

cc: @dmikusa-pivotal 

Maybe blocked on https://github.com/buildpacks/spec/pull/193/files but we already support k8s service bindings during runtime which is currently not defined in the spec.

### DIFF
--- a/platform.go
+++ b/platform.go
@@ -35,6 +35,18 @@ const (
 
 	// BindingType is the key for a binding's type.
 	BindingType = "type"
+
+	// EnvServiceBindings is the name of the environment variable that contains the path to service bindings directory.
+	//
+	// See the Service Binding Specification for Kubernetes for more details - https://k8s-service-bindings.github.io/spec/
+	EnvServiceBindings = "SERVICE_BINDING_ROOT"
+
+	// EnvCNBBindings is the name of the environment variable that contains the path to the CNB bindings directory. The CNB
+	// bindings spec will eventually by deprecated in favor of the Service Binding Specification for Kubernetes -
+	// https://github.com/buildpacks/rfcs/blob/main/text/0055-deprecate-service-bindings.md.
+	//
+	// See the CNB bindings extension spec for more details - https://github.com/buildpacks/spec/blob/main/extensions/bindings.md
+	EnvCNBBindings = "CNB_BINDINGS"
 )
 
 // Binding is a projection of metadata about an external entity to be bound to.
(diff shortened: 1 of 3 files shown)


dmikusa-pivotal:
Awesome, thanks for writing this up! 

+1 for changes
+1 for not waiting on the builder extension spec
  a.) it would be nice to have this ASAP, as there are docs that reference using this env variable that don't work at the moment
  b.) it doesn't seem like the spec is going to change anything, just ratify what is being done here. If something does change, we can always adjust libcnb to that at a later date, but we have it working now.

platform.go
```diff
@@ -167,6 +167,21 @@ func NewBindingsFromPath(path string) (Bindings, error) {
...
+		return NewBindingsFromPath(path)
+	}
+
+	// TODO: Remove as CNB_BINDINGS ages out
+	if path, ok := os.LookupEnv("CNB_BINDINGS"); ok {
+		return NewBindingsFromPath(path)
+	}
+
+	return NewBindingsFromPath(path)
+}
```
ekcasey, line 183:
```suggestion
// NewBindingsForBuild creates a new bindings from all the bindings at the path defined by $SERVICE_BINDING_ROOT
// or $CNB_BINDINGS if it does not exist.  If neither is defined, bindings are read from <platform>/bindings, the default
// path defined in the CNB Binding extension specification.
func NewBindingsForBuild(platformDir string) (Bindings, error) {
	if path, ok := os.LookupEnv("SERVICE_BINDING_ROOT"); ok {
		return NewBindingsFromPath(path)
	}

	// TODO: Remove as CNB_BINDINGS ages out
	if path, ok := os.LookupEnv("CNB_BINDINGS"); ok {
		return NewBindingsFromPath(path)
	}

	return NewBindingsFromPath(filepath.Join(platformDir, "bindings"))
}
```
I think we should do this ^ and rename `NewBindingsFromEnvironment` to `NewBindingsForLaunch` or similar to make it clear to users that the default path is only defined in the build environment. I also think it is nice to move the default path calculation into the helper function so all of the boilerplate for finding the bindings dir is in a single place.

platform.go
```diff
@@ -167,6 +167,21 @@ func NewBindingsFromPath(path string) (Bindings, error) {
 	return bindings, nil
 }
 
+// NewBindingsFromEnvOrPath creates a new bindings from all the bindings at the path defined by $SERVICE_BINDING_ROOT
+// or $CNB_BINDINGS if it does not exist.  If neither is defined, it defaults to using the given path.
+func NewBindingsFromEnvOrPath(path string) (Bindings, error) {
+	if path, ok := os.LookupEnv("SERVICE_BINDING_ROOT"); ok {
```
ekcasey, line 173: It could be nice to extract `SERVICE_BINDING_ROOT` and `CNB_BINDINGS` into exported constants like `EnvServiceBindingRoot` etc.

platform.go
samj1912: Renamed `NewBindingsFromEnvironment` but kept the old one around since it is still used.

samj1912:
@buildpacks/buildpack-authors-tooling-maintainers ping :)

ekcasey (approved):

PR #63, opened 2021-06-28
Add a makefile to the repository to run linting and tests

### LINKED ISSUE #61: Add Makefile
Following the conventions of other buildpacks repos, we should add a Makefile to this repo that runs tests and static checks.

samj1912:
Fixes #61


Makefile
```diff
@@ -0,0 +1,25 @@
...
+install-golangci-lint:
+	@echo "> Installing golangci-lint..."
+	cd tools && $(GOCMD) install github.com/golangci/golangci-lint/cmd/golangci-lint
+
+lint: install-golangci-lint
+	@echo "> Linting code..."
+	@golangci-lint run -c golangci.yaml
+
+test: format lint
+	$(GOCMD) test -parallel=1 -count=1 -v ./...
```
ekcasey, line 25:
```suggestion
	$(GOCMD) test -parallel=1 -count=1 -v ./...

```

golangci.yaml
```diff
@@ -0,0 +1,32 @@
+run:
+  timeout: 6m
+
+linters:
+  disable-all: true
+  enable:
+    - bodyclose
+    - deadcode
+    - dogsled
```
ekcasey, line 9:
```suggestion
    - dogsled
    - exportloopref
```

Replaces `scopelint`:
```
WARN [runner] The linter 'scopelint' is deprecated (since v1.39.0) due to: The repository of the linter has been deprecated by the owner.  Replaced by exportloopref.
```

golangci.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - gocritic
+    - goimports
+    - golint
```
ekcasey, line 12:
```suggestion
```

replaced by `revive`:
```
WARN [runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive.
richgo test -parallel=1 -count=1 -v ./...
```

golangci.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - gosec
+    - gosimple
+    - govet
+    - ineffassign
+    - maligned
```
ekcasey, line 17:
```suggestion
```

handled by `govet`:
```
WARN [runner] The linter 'maligned' is deprecated (since v1.38.0) due to: The repository of the linter has been archived by the owner.  Replaced by govet 'fieldalignment'.
```

golangci.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - misspell
+    - nakedret
```
ekcasey, line 19:
```suggestion
    - nakedret
    - revive
```

Replaces `golint`, see above.

golangci.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - scopelint
```
ekcasey, line 20:
```suggestion
```
Replaced by `exportloopref`. See above.

golangci.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - stylecheck
+    - typecheck
+    - unconvert
+    - unused
+    - varcheck
+    - whitespace
+
+linters-settings:
+  goimports:
+    local-prefixes: github.com/buildpacks/libcnb
```
ekcasey, line 32:
```suggestion
    local-prefixes: github.com/buildpacks/libcnb

```

ekcasey (changes requested):

.github/workflows/tests.yml
```diff
@@ -44,6 +44,6 @@ jobs:
 
                 set -euo pipefail
 
-                richgo test ./...
+                GOCMD=richgo make
```
ekcasey, line 47: Sorry I lost this comment on the original review. This tests are managed by https://github.com/paketo-buildpacks/pipeline-builder. In order to prevent the pipeline builder from continually trying to undo this change, we need to [configure `test`]rhttps://github.com/paketo-buildpacks/pipeline-builder#test) in https://github.com/buildpacks/libcnb/blob/main/.github/pipeline-descriptor.yml. Or, move away from pipeline builder entirely.

ekcasey: Or convince @dmikusa-pivotal to add a `Makefile` to all the Paketo buildpacks that use pipeline builder and update the default `test` config ;). Which seems like a good idea honestly, but is probably a lot to ask him to do today.

samj1912: Oh wow - did not know about this. Added the configuration for now :P I don't wanna put more on @dmikusa-pivotal's plate.

dmikusa-pivotal: I wouldn't be opposed to parameterizing it. It would be a small change, something like `TEST_CMD`. The template it creates would run `$TEST_CMD` then. I could also see parameterizing GOCMD, not everyone might want to use `richgo`. Then you wouldn't need to set GOCMD, it would just be set by the workflow based on the pipeline descriptor values. If that seems reasonable, I can open an issue to make the change.

dmikusa-pivotal:
@samj1912 Out of curiosity, why did all of the file permissions change in the tests? Is that an artifact of running with make? Not that it seems like a problem to change them, it just makes this PR a lot larger.

samj1912:
> @samj1912 Out of curiosity, why did all of the file permissions change in the tests? Is that an artifact of running with make? Not that it seems like a problem to change them, it just makes this PR a lot larger.

That was due to the lint configuration (gosec). I separated those changes from the makefile changes in different commits.

.github/pipeline-descriptor.yml
```diff
@@ -5,3 +5,22 @@ github:
...
+      GO111MODULE=on go get -u -ldflags="-s -w" github.com/kyoh86/richgo
+  - name: Run Tests
+    run: |
+      #!/usr/bin/env bash
+
+      set -euo pipefail
+
+      GOCMD=richgo make
+    env:
+      RICHGO_FORCE_COLOR: "1"
```
ekcasey, line 26:
```suggestion
      RICHGO_FORCE_COLOR: "1"

```

.github/workflows/tests.yml
ekcasey: > I wouldn't be opposed to parameterizing it. It would be a small change, something like TEST_CMD.

Personally I think overriding the whole step is fine in this situation.

ekcasey (approved):

.github/workflows/tests.yml
dmikusa-pivotal: Sorry @ekcasey, I'm not sure I follow what you mean. It's OK though, if you or @samj1912 want to submit an issue with details of what you need to https://github.com/paketo-buildpacks/pipeline-builder/issues, I'll make it happen for you.

samj1912:
@ekcasey I don't have merge permissions.

PR #74, opened 2021-09-22
Removes LayerContributor concept

ryanmoran:
The buildpack author tooling subteam has been having an ongoing discussion around making some changes to `libcnb` that would allow it to be more flexible and thus more consumable by different buildpack development teams. While the abstract conversations have been good to get started, I feel like having a concrete implementation we can start to move forward with might also be helpful.

One of the items of discussion has been replacing the `LayerContributor` concept with a simpler `Layer` API that is similar to what [`packit`](https://github.com/paketo-buildpacks/packit) provides. This PR removes `LayerContributor` and instead expects that a `BuildResult` will contain a slice of `Layer` values once the build phase of the buildpack completes.

The change in expectation here is that any modification to the layer contents should be handled by the buildpack itself rather than delegated to be invoked by `libcnb` as the current API requires. While the `LayerContributor` is removed, it is still relatively easy to make a change to this new API without replacing a buildpack's contributors entirely. Buildpacks can simply call the `Contribute` methods on their contributors prior to returning the `BuildResult`.


samj1912:
@ryanmoran this seems to be a backwards incompatible change. Can we instead add a new param `OutputLayers` and add a deprecation notice on `Layers`?

dmikusa-pivotal:
I'm not opposed to just removing it. I think that could make the API cleaner. If we add `OutputLayers` that's more verbose than just `Layers`, and then that name is locked into the API. Not the end of the world, but maybe it's time to start dev work on v2 now.

We could:

1. Add this PR as part of a new v2 branch where we can make breaking API changes
2. Continue supporting the v1 branch while the development of v2 progresses & for say 6 months after. That gives users time to migrate off.

This would also give us the opportunity to clean up other facets of the API or make other changes that we've been discussing in the WG.

ryanmoran:
I'm aware that this is a backwards-incompatible change. I don't expect this to be merged, at least not directly to `main`. The point is to try to get the conversation to something more concrete around a real API. I agree with @dmikusa-pivotal that we should start to think about what a `v2` should look like. This isn't the only change I'd like to see in `libcnb` before we'd be able to adopt it more completely in Paketo.

samj1912 (approved):

samj1912:
LGTM - I have created a branch called `version-2.0` to stack on changes for libcnb v2 (didn't want to create a branch called v2 as that may have conflicted with the tag later on). Thanks for kicking this off!

PR #75, opened 2021-09-23
Added logging of bindings & fixed bindings bug 

pivotal-david-osullivan:
Fixes a bug where Bindings are silently created from hidden dirs. These were ignored when ConfigMap were being created but not for Bindings themselves.

This PR also logs bindings as pairs of 'name & keys' that are identified during the build, since it can be difficult to know whether bindings have been found as expected during the build process. It also prints a line to indicate if no bindings were identified, for the same reason.


build.go
```diff
@@ -183,7 +185,7 @@ func Build(builder Builder, options ...Option) {
 		config.exitHandler.Error(fmt.Errorf("unable to read platform bindings %s\n%w", ctx.Platform.Path, err))
 		return
 	}
-	logger.Debugf("Platform Bindings: %+v", ctx.Platform.Bindings)
+	numBindings = len(ctx.Platform.Bindings)
```
dmikusa-pivotal, line 188: It seems like you could probably move this into the `logBindings` method. I don't see it used outside of that function.

dmikusa-pivotal (approved):

dmikusa-pivotal:
For what it's worth, the driving factor behind this is that people really struggle with bindings. In many cases, users don't get the binding format correct, but the buildpack doesn't output anything in either case so it's not clear if a buildpack sees or does not see them. This makes debugging difficult.

This change moves them front and center, so it's obvious when they're being read by a buildpack which makes debugging easier.

build.go
pivotal-david-osullivan: 👍

samj1912 (dismissed):

build.go
```diff
@@ -364,6 +364,34 @@ func Build(builder Builder, options ...Option) {
...
+			logger.Infof("  Name: %s", binding.Name)
+			if len(s) > 0 {
+				logger.Infof("    Keys: %s", strings.Join(s, ", "))
+			} else {
+				logger.Infof("    Keys: None")
+			}
+			s = nil
+		}
+	} else {
+		logger.Infof("No Platform Bindings Identified")
```
samj1912, line 393:
```suggestion
		logger.Info("No Platform Bindings Identified")
```

build.go
```diff
@@ -364,6 +364,34 @@ func Build(builder Builder, options ...Option) {
...
+		for _, binding := range bindings {
+			for k := range binding.Secret {
+				s = append(s, k)
+			}
+			sort.Strings(s)
+			logger.Infof("  Name: %s", binding.Name)
+			if len(s) > 0 {
+				logger.Infof("    Keys: %s", strings.Join(s, ", "))
+			} else {
+				logger.Infof("    Keys: None")
```
samj1912, line 388:
```suggestion
				logger.Info("    Keys: None")
```

build.go
```diff
@@ -364,6 +364,34 @@ func Build(builder Builder, options ...Option) {
 			return
 		}
 	}
+	logBindings(ctx.Platform.Bindings, logger)
```
samj1912, line 367: Curious why we are logging this at the end instead of L#186?

build.go
pivotal-david-osullivan: 👍

build.go
pivotal-david-osullivan: I had the logging here initially but in testing, the bindings appeared before the buildpack title & output which seemed confusing. Here it appears at the end of the relevant buildpacks output.

build.go
```diff
@@ -364,6 +364,34 @@ func Build(builder Builder, options ...Option) {
...
+}
+
+func logBindings(bindings Bindings, logger poet.Logger) {
+	numBindings := len(bindings)
+	if numBindings > 0 {
+		if numBindings == 1 {
+			logger.Info("\n1 Binding Identified:")
```
ekcasey, line 374: Should this be debug level logging? It seems like it could be annoying if you add a binding intended to pass data to a single buildpack and every buildpack in the build lists every binding.

Also it seems like we typically log at the debug level in libcnb unless something goes wrong, allowing the buildpacks themselves to design the default output

pivotal-david-osullivan:
Closing in favour of #79 which only fixes the bug mentioned in this PR

PR #78, opened 2021-10-02
Add execd helper

samj1912:
This adds functionality in libcnb to define execd interfaces easily.


exec_d.go
```diff
@@ -0,0 +1,71 @@
...
+	"path/filepath"
+
+	"github.com/buildpacks/libcnb/internal"
+)
+
+//go:generate mockery --name Executor --case=underscore
+
+// Executor describes an interface for types that follow the Exec.d specification.
+// It should return a map of environment variables and their values as output.
+type Executor interface {
```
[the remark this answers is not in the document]
dmikusa-pivotal: Or possibly having the interface be `ExecD`, which lines up with the spec pretty well too, and having the `ExecD` function be something different like `InvokeExecD` or `RunExecD`? Since the primary purpose of that function is to trigger the setup & execution of an ExecD executable.

dmikusa-pivotal (changes requested):
Not necessarily something we have to change, but I wanted to start a conversation about the naming of the primary type & function this change exposes.

I outlined a couple of options that I think might follow the terminology used in the spec a little more closely. IMHO, that would help someone trying to use the library as it's easy to see how the library fits with the spec.

exec_d.go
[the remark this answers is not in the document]
samj1912: Not too tied to the existing names. I was also struggling with the appropriate names for these. I like the latter idea since it makes it clear what's happening but also okay with the first one.

samj1912, line 31:
```suggestion
type ExecD interface {
```

exec_d.go
```diff
@@ -0,0 +1,71 @@
...
+	Execute() (map[string]string, error)
+}
+
+// ExecD is called by the main function of a buildpack's execd binary, encompassing multiple execd
+// executors in one binary.
+func ExecD(executors map[string]Executor, options ...Option) {
```
samj1912, line 37:
```suggestion
func RunExecD(execdMap map[string]ExecD, options ...Option) {
```

samj1912 (commented):
@dmikusa-pivotal just confirming if the suggested names look okay.

exec_d.go
dmikusa-pivotal: 👍

exec_d.go
dmikusa-pivotal: 👍

dmikusa-pivotal (approved):

PR #86, opened 2021-11-04
Add struct tag to Buildpack.Path field

### LINKED ISSUE #85: Buildpack struct's Path field missing struct tag
When serializing a `Buildpack` instance to TOML, the `Path` field is serialized as `Path = "..."` regardless of whether or not it is set. The capital `P` in `Path` makes this serialized TOML out-of-spec. I will be submitting a PR to fix this. Thanks!

jghiloni:
Most Go applications use `github.com/burntsushi/toml` as their toml parsing library. When struct tags are omitted, the library will do case-insensitive deserialization, so a property called `path` will successfully unmarshal into the `Path` struct field. 

However, if youmarshal a Buildpack instance to toml, the `Path` field appears regardless of whether it is set or not, and technically violates the buildpack toml spec. This PR adds a struct tag to not only marshal the field to the `path` property in the resultant toml, but also adds the `omitempty` modifier to ignore the field altogether when not set.

This resolves #85 

Signed-off-by: Josh Ghiloni <jghiloni@vmware.com>

### DIFF
--- a/buildpack_test.go
+++ b/buildpack_test.go
@@ -45,6 +45,6 @@ func testBuildpackTOML(t *testing.T, context spec.G, it spec.S) {
 		output := &bytes.Buffer{}
 
 		Expect(toml.NewEncoder(output).Encode(bp)).To(Succeed())
-		Expect(output.String()).NotTo(ContainSubstring("ath = ")) // match on path and Path
+		Expect(output.String()).NotTo(Or(ContainSubstring("Path = "), ContainSubstring("path = ")))
 	})
 }
--- a/buildpack_test.go
+++ b/buildpack_test.go
@@ -32,22 +32,7 @@ func testBuildpackTOML(t *testing.T, context spec.G, it spec.S) {
 		Expect = NewWithT(t).Expect
 	)
 
-	it("does not serialize an empty Path field", func() {
-		bp := libcnb.Buildpack{
-			API: "0.6",
-			Info: libcnb.BuildpackInfo{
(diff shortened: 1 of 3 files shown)


dmikusa-pivotal:
I don't believe that `path` is part of the spec for `buildpack.toml`. My understanding is that [libcnb adds this field so an author knows where the buildpack resides](https://github.com/buildpacks/libcnb/blob/e90023ce444ec12c176fbcb050fd07087f8f4afb/build.go#L152-L156). I think we should just omit that field altogether if we're reading or writing a buildpack.toml.

jghiloni:
That sounds good -- I'll add another commit changing that behavior

dmikusa-pivotal (approved):

buildpack_test.go
```diff
@@ -0,0 +1,50 @@
...
+				ID:   "test-buildpack/sample",
+				Name: "sample",
+			},
+			Path: "../buildpack",
+		}
+
+		output := &bytes.Buffer{}
+
+		Expect(toml.NewEncoder(output).Encode(bp)).To(Succeed())
+		Expect(output.String()).NotTo(ContainSubstring("ath = ")) // match on path and Path
```
samj1912, line 48: Can we use the `Or` matcher inside `Not` and match exactly for Path and path?

jghiloni: Yeah, that makes sense! I'll add something.

samj1912 (approved):

PR #89, opened 2021-11-05
Renames poet -> log

### LINKED ISSUE #81: libcnb v2: Rename poet to logging
Right now libcnb uses the package name of `poet` for its logging package.

https://github.com/buildpacks/libcnb/tree/main/poet

We should rename this to something more boring and straightforward so that it's easier to see what the package does and more accessible for non-native English speakers.

Presently you would use the library in a manner like this:

```
logger := poet.NewLogger(os.Stdout)
```

given this, other options would be:

1. `logger := logging.NewLogger(os.Stdout)`
2. `logger := log.NewLogger(os.Stdout)`
3. `logger := logs.NewLogger(os.Stdout)`

Please add your thoughts on these options below or other suggestions for package names.

dmikusa-pivotal:
- Renames the poet module to log
- Renames NewLogger to just New, so that you have less redundancy. i.e. log.NewLogger becomes log.New
- Renames NewLoggerWithOptions to just NewWithOptions, so that you have less redundancy. i.e. log.NewLoggerWithOptions becomes log.NewWithOptions
- Renames LogLevel to just Level, so that you have less redundancy. i.e. log.LogLevel becomes log.Level

Resolves #81

### DIFF
--- a/build.go
+++ b/build.go
@@ -27,7 +27,7 @@ import (
 	"github.com/BurntSushi/toml"
 
 	"github.com/buildpacks/libcnb/internal"
-	"github.com/buildpacks/libcnb/poet"
+	"github.com/buildpacks/libcnb/log"
 )
 
 // BuildContext contains the inputs to build.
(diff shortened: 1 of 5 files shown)


dmikusa-pivotal:
This isn't the PR I mentioned in the WG meeting today, but probably a first step. It resolves #89.

samj1912 (approved):

PR #95, opened 2021-11-16
Adds a convenience method for getting the build, launch and layer BOM file paths

dmikusa-pivotal:
Also adds `Deprecated:` godoc comments.

### DIFF
--- a/application.go
+++ b/application.go
@@ -73,6 +73,8 @@ type LaunchTOML struct {
 	Slices []Slice `toml:"slices"`
 
 	// BOM is a collection of entries for the bill of materials.
+	//
+	// Deprecated: as of Buildpack API 0.7, write to `layer.BOMPath()` instead
 	BOM []BOMEntry `toml:"bom"`
 }
 
@@ -83,6 +85,8 @@ func (l LaunchTOML) isEmpty() bool {
 // BuildTOML represents the contents of build.toml.
 type BuildTOML struct {
 	// BOM contains the build-time bill of materials.
+	//
+	// Deprecated: as of Buildpack API 0.7, write to `layer.BOMPath()` instead
 	BOM []BOMEntry `toml:"bom"`
 
 	// Unmet is a collection of buildpack plan entries that should be passed through to subsequent providers.
@@ -94,6 +98,8 @@ func (b BuildTOML) isEmpty() bool {
 }
 
 // BOMEntry contains a bill of materials entry.
+//
+// Deprecated: as of Buildpack API 0.7, BOM should use standard formats like CycloneDX going forward
 type BOMEntry struct {
 	// Name represents the name of the entry.
 	Name string `toml:"name"`
--- a/layer.go
+++ b/layer.go
@@ -28,7 +28,6 @@ import (
 
 // Exec represents the exec.d layer location
 type Exec struct {
-
 	// Path is the path to the exec.d directory.
 	Path string
 }
@@ -68,9 +67,21 @@ func (p Profile) ProcessAddf(processType string, name string, format string, a .
 	p.Addf(filepath.Join(processType, name), format, a...)
 }
 
-// Contribute represents a layer managed by the buildpack.
-type Layer struct {
+// BOMFormat indicates the format of the SBOM entry
+type SBOMFormat int
 
+const (
+	CycloneDXJSON SBOMFormat = iota
+	SPDXJSON
+	SyftJSON
+)
+
+func (b SBOMFormat) String() string {
+	return []string{"cdx.json", "spdx.json", "syft.json"}[b]
+}
+
+// Layer represents a layer managed by the buildpack.
+type Layer struct {
 	// LayerTypes indicates the type of layer
 	LayerTypes `toml:"types"`
 
@@ -99,6 +110,10 @@ type Layer struct {
 	Exec Exec `toml:"-"`
 }
 
+func (l Layer) SBOMPath(bt SBOMFormat) string {
+	return filepath.Join(filepath.Dir(l.Path), fmt.Sprintf("%s.sbom.%s", l.Name, bt))
+}
+
 // LayerTypes describes which types apply to a given layer. A layer may have any combination of Launch, Build, and
 // Cache types.
 type LayerTypes struct {
@@ -116,7 +131,6 @@ type LayerTypes struct {
 
 // LayerContributor is an interface for types that create layers.
 type LayerContributor interface {
-
 	// Contribute accepts a layer and transforms it, returning a layer.
 	Contribute(layer Layer) (Layer, error)
 
@@ -126,7 +140,6 @@ type LayerContributor interface {
 
 // Layers represents the layers part of the specification.
 type Layers struct {
-
 	// Path is the layers filesystem location.
 	Path string
 }
@@ -161,4 +174,14 @@ func (l *Layers) Layer(name string) (Layer, error) {
 	}
 
 	return layer, nil
+}
+
+// BOMBuildPath returns the full path to the build SBoM file for the buildpack
+func (l Layers) BuildSBOMPath(bt SBOMFormat) string {
+	return filepath.Join(l.Path, fmt.Sprintf("build.sbom.%s", bt))
+}
+
+// BOMLaunchPath returns the full path to the launch SBoM file for the buildpack
+func (l Layers) LaunchSBOMPath(bt SBOMFormat) string {
+	return filepath.Join(l.Path, fmt.Sprintf("launch.sbom.%s", bt))
 }
--- a/layer_test.go
+++ b/layer_test.go
@@ -110,6 +110,18 @@ func testLayer(t *testing.T, context spec.G, it spec.S) {
 			Expect(l.Profile).To(Equal(libcnb.Profile{}))
 		})
 
+		it("generates BOM paths", func() {
+			l, err := layers.Layer("test-name")
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(l.Path).To(Equal(filepath.Join(path, "test-name")))
+			Expect(layers.BuildSBOMPath(libcnb.CycloneDXJSON)).To(Equal(filepath.Join(path, "build.sbom.cdx.json")))
+			Expect(layers.BuildSBOMPath(libcnb.SPDXJSON)).To(Equal(filepath.Join(path, "build.sbom.spdx.json")))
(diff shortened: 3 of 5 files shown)


layer.go
```diff
@@ -68,9 +67,46 @@ func (p Profile) ProcessAddf(processType string, name string, format string, a .
...
+type BOMLocation int
+
+const (
+	BuildBOM BOMLocation = iota
+	LaunchBOM
+	LayerBOM
+)
+
+// String provides a textual representation
+func (b BOMLocation) String() string {
```
samj1912, line 93: When would this be useful? Also is the `LayerBOM` mapping to `<layer>` potentially confusing?

layer.go
```diff
@@ -99,6 +135,21 @@ type Layer struct {
...
+	switch bl {
+	case BuildBOM:
+		bom = "build"
+	case LaunchBOM:
+		bom = "launch"
+	case LayerBOM:
+		bom = l.Name
+	}
+
+	return filepath.Join(filepath.Dir(l.Path), fmt.Sprintf("%s.bom.%s", bom, bt))
```
samj1912, line 150:
```suggestion
	return filepath.Join(filepath.Dir(l.Path), fmt.Sprintf("%s.sbom.%s", bom, bt))
```

layer_test.go
```diff
@@ -110,6 +110,18 @@ func testLayer(t *testing.T, context spec.G, it spec.S) {
...
+		it("generates BOM paths", func() {
+			l, err := layers.Layer("test-name")
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(l.Path).To(Equal(filepath.Join(path, "test-name")))
+			Expect(l.BOMPath(libcnb.BuildBOM, libcnb.CycloneDXJSON)).To(Equal(filepath.Join(path, "build.bom.cdx.json")))
+			Expect(l.BOMPath(libcnb.BuildBOM, libcnb.SPDXJSON)).To(Equal(filepath.Join(path, "build.bom.spdx.json")))
+			Expect(l.BOMPath(libcnb.BuildBOM, libcnb.SyftJSON)).To(Equal(filepath.Join(path, "build.bom.syft.json")))
+			Expect(l.BOMPath(libcnb.LaunchBOM, libcnb.SyftJSON)).To(Equal(filepath.Join(path, "launch.bom.syft.json")))
+			Expect(l.BOMPath(libcnb.LayerBOM, libcnb.SyftJSON)).To(Equal(filepath.Join(path, "test-name.bom.syft.json")))
```
samj1912, line 122:
```suggestion
			Expect(l.Path).To(Equal(filepath.Join(path, "test-name")))
			Expect(l.BOMPath(libcnb.BuildBOM, libcnb.CycloneDXJSON)).To(Equal(filepath.Join(path, "build.sbom.cdx.json")))
			Expect(l.BOMPath(libcnb.BuildBOM, libcnb.SPDXJSON)).To(Equal(filepath.Join(path, "build.sbom.spdx.json")))
			Expect(l.BOMPath(libcnb.BuildBOM, libcnb.SyftJSON)).To(Equal(filepath.Join(path, "build.sbom.syft.json")))
			Expect(l.BOMPath(libcnb.LaunchBOM, libcnb.SyftJSON)).To(Equal(filepath.Join(path, "launch.sbom.syft.json")))
			Expect(l.BOMPath(libcnb.LayerBOM, libcnb.SyftJSON)).To(Equal(filepath.Join(path, "test-name.sbom.syft.json")))
```

samj1912 (changes requested):

layer.go
dmikusa-pivotal: I had originally added this with the intent that it would help for logging and error messages, i.e. `%s` would resolve to the textual representation here.

I didn't end up using this in the libpak implementation, so if you're opposed to having it, I'm OK with removing it. 

With that in mind, I opted to have `<layer>` because you can't know the actual value without having access to the layer to which the location applies (since it would be the layer name). This seemed good enough for logging/error messages, but if you feel we should just drop it, I'm OK with that.

dmikusa-pivotal: My latest commit removes this. It seems to be of limited value and potential for confusion, so it's out. We can add something later if there's a demonstrated usecase.

layer.go
```diff
@@ -68,9 +67,30 @@ func (p Profile) ProcessAddf(processType string, name string, format string, a .
 	p.Addf(filepath.Join(processType, name), format, a...)
 }
 
-// Contribute represents a layer managed by the buildpack.
-type Layer struct {
+// BOMFormat indicates the format of the SBOM entry
+type BOMFormat int
+
+const (
+	CycloneDXJSON BOMFormat = iota
```
samj1912, line 74: This implies that a default of cyclonedx would be picked if not specified. Do we instead want this to be undefined and detect and error if someone tries to access that?

dmikusa-pivotal: I'm not sure I totally follow what you're suggestion in regards to making it undefined. Are you suggesting we add a value to the enum that is Undefined and make it the first element? I'm not very keen on that. Or is there something else I'm missing?

My present thought is that I'm fine with it being the default. Not to imply it's a better or more important format than any other format. It is just alphabetically first, so why not?

layer.go
```diff
@@ -99,6 +119,21 @@ type Layer struct {
...
 }
 
+func (l Layer) BOMPath(bl BOMLocation, bt BOMFormat) string {
+	var bom string
+
+	switch bl {
+	case BuildBOM:
+		bom = "build"
+	case LaunchBOM:
+		bom = "launch"
```
samj1912, line 129: It seems a bit weird to me that the buildpack level boms would be attached to the layer. IMHO this should only return the layer specific bom path and the build contexts or the `Layers` object should contain the top level bom files.

dmikusa-pivotal: I can take a look at changing this. It ended up being pretty convenient when I was testing this out. Due to the layer contributor approach, you usually have a layer when you're going to write the SBoM. I can think of at least one case where this might not be true though, so let me play around a little more.

Excellent feedback, keep it coming :)

dmikusa-pivotal: OK, I can work with that in libpak. I submitted an update with this change. It moves build and launch SBoM methods to the Layers object.

dmikusa-pivotal:
@samj1912 - Do you want me to squash all the commits? Or will you do that on merge?

samj1912:
Squash would be nice! As a project we try to avoid changing commits from the authors and use "merge" instead.

dmikusa-pivotal:
👍 @samj1912 - squashed.

layer.go
```diff
@@ -162,3 +184,13 @@ func (l *Layers) Layer(name string) (Layer, error) {
 
 	return layer, nil
 }
+
+// BOMBuildPath returns the full path to the build SBoM file for the buildpack
+func (l Layers) BOMBuildPath(bt BOMFormat) string {
```
samj1912, line 189:
```suggestion
func (l Layers) BuildSBOMPath(bt SBOMFormat) string {
```

layer.go
```diff
@@ -162,3 +184,13 @@ func (l *Layers) Layer(name string) (Layer, error) {
...
+	return filepath.Join(l.Path, fmt.Sprintf("build.sbom.%s", bt))
+}
+
+// BOMLaunchPath returns the full path to the launch SBoM file for the buildpack
+func (l Layers) BOMLaunchPath(bt BOMFormat) string {
```
samj1912, line 194:
```suggestion
func (l Layers) LaunchSBOMPath(bt SBOMFormat) string {
```

layer.go
```diff
@@ -99,6 +119,10 @@ type Layer struct {
 	Exec Exec `toml:"-"`
 }
 
+func (l Layer) BOMPath(bt BOMFormat) string {
```
samj1912, line 122:
```suggestion
func (l Layer) SBOMPath(bt SBOMFormat) string {
```

layer.go
```diff
@@ -68,9 +67,30 @@ func (p Profile) ProcessAddf(processType string, name string, format string, a .
...
+}
+
+// BOMLocation indicates the location of the SBOM entry
+type BOMLocation int
+
+const (
+	BuildBOM BOMLocation = iota
+	LaunchBOM
+	LayerBOM
+)
```
samj1912, line 90: We shouldn't need this anymore.

layer.go
```diff
@@ -68,9 +67,30 @@ func (p Profile) ProcessAddf(processType string, name string, format string, a .
...
+// BOMFormat indicates the format of the SBOM entry
+type BOMFormat int
+
+const (
+	CycloneDXJSON BOMFormat = iota
+	SPDXJSON
+	SyftJSON
+)
+
+func (b BOMFormat) String() string {
```
samj1912, line 79:
```suggestion
func (b SBOMFormat) String() string {
```

layer.go
```diff
@@ -68,9 +67,30 @@ func (p Profile) ProcessAddf(processType string, name string, format string, a .
 	p.Addf(filepath.Join(processType, name), format, a...)
 }
 
-// Contribute represents a layer managed by the buildpack.
-type Layer struct {
+// BOMFormat indicates the format of the SBOM entry
+type BOMFormat int
```
samj1912, line 71:
```suggestion
type SBOMFormat int
```

layer.go
samj1912, line 74:
```suggestion
	CycloneDXJSON SBOMFormat = iota
```

samj1912 (changes requested):
Just a few changes for the sake of consistency

layer.go
dmikusa-pivotal: Weird. I thought I had removed this. Definitely removed now.

dmikusa-pivotal:
> Just a few changes for the sake of consistency

All set.

application.go
```diff
@@ -73,6 +73,8 @@ type LaunchTOML struct {
 	Slices []Slice `toml:"slices"`
 
 	// BOM is a collection of entries for the bill of materials.
+	//
+	// Deprecated: as of RFC 95, write to `layer.BOMPath()` instead
```
samj1912, line 77:
```suggestion
	// Deprecated: as of Buildpack API 0.7, write to `layer.BOMPath()` instead
```

samj1912: Same for other mentions of RFC 95.

samj1912 (approved):
Minor comment but apart from that LGTM

application.go
dmikusa-pivotal: Updated.

samj1912:
Thanks for the PR!

PR #97, opened 2021-11-19
Additional updates for SBOM Support

dmikusa-pivotal:
Not required to use the new SBOM support, but here are some additional helpful bits.

- Adds SBOMFormats (maps to `sbom-formats` in buildpack.toml) to the BuildpackInfo struct. This makes the information accessible to buildpacks.
- Modifies build such that it only writes the old-style build and launch BOM information if the buildpack API is less than 0.7. If it's 0.7+, it should not write the old-style format as that can conflict with the new SBOM format and cause the lifecycle to fail.
- Adds validation of the new SBOM files that are written by a buildpack. We check that the extension matches up with a valid MIME type that is listed in `buildpack.toml`'s `sbom-formats` field. If it does not match up, then it fails. This should not generally happen with published buildpacks. This check can be helpful while authoring buildpacks, to ensure everything is correctly set up.

### DIFF
--- a/build.go
+++ b/build.go
@@ -306,6 +306,11 @@ func Build(builder Builder, options ...Option) {
 		}
 	}
 
+	if err := ValidateSBOMFormats(ctx.Layers.Path, ctx.Buildpack.Info.SBOMFormats); err != nil {
+		config.exitHandler.Error(fmt.Errorf("unable to validate SBOM\n%w", err))
+		return
+	}
+
 	// Deprecated: as of Buildpack API 0.7, to be removed in a future version
 	var launchBOM, buildBOM []BOMEntry
 	if result.BOM != nil {
@@ -338,6 +343,12 @@ func Build(builder Builder, options ...Option) {
 			}
 		}
 
+		// even if there is data, do not write a BOM if we have buildpack API 0.7, that will cause a lifecycle error
+		if API == "0.7" {
+			logger.Info("Warning: this buildpack is including both old and new format SBOM information, which is an invalid state. To prevent the lifecycle from failing, libcnb is discarding the old SBOM information.")
+			launch.BOM = nil
+		}
+
 		if err = config.tomlWriter.Write(file, launch); err != nil {
 			config.exitHandler.Error(fmt.Errorf("unable to write application metadata %s\n%w", file, err))
 			return
@@ -352,6 +363,13 @@ func Build(builder Builder, options ...Option) {
 	if !build.isEmpty() {
 		file = filepath.Join(ctx.Layers.Path, "build.toml")
 		logger.Debugf("Writing build metadata: %s <= %+v", file, build)
+
+		// even if there is data, do not write a BOM if we have buildpack API 0.7, that will cause a lifecycle error
+		if API == "0.7" {
+			logger.Info("Warning: this buildpack is including both old and new format SBOM information, which is an invalid state. To prevent the lifecycle from failing, libcnb is discarding the old SBOM information.")
+			build.BOM = nil
+		}
+
 		if err = config.tomlWriter.Write(file, build); err != nil {
 			config.exitHandler.Error(fmt.Errorf("unable to write build metadata %s\n%w", file, err))
 			return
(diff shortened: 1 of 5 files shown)


build.go
```diff
@@ -380,3 +396,27 @@ func contains(candidates []string, s string) bool {
 
 	return false
 }
+
+func ValidateSBOMFormats(layersPath string, acceptedSBOMFormats []string) error {
+	sbomFiles, err := filepath.Glob(fmt.Sprintf("%s/*.sbom.*", layersPath))
```
samj1912, line 401: Can we use filepath join here to combine the glob with the layer path?

layer.go
```diff
@@ -74,10 +74,28 @@ const (
...
 func (b SBOMFormat) String() string {
-	return []string{"cdx.json", "spdx.json", "syft.json"}[b]
+	return []string{"cdx.json", "spdx.json", "syft.json", "unknown"}[b]
+}
+
+func (b SBOMFormat) MimeType() string {
+	return []string{"application/vnd.cyclonedx+json", "application/spdx+json", "application/vnd.syft+json", "unknown"}[b]
+}
+
+func SBOMFormatFromString(from string) (SBOMFormat, error) {
```
samj1912, line 88: Should we create a slice of all known sbom formats? That way we can just iterate over them.

This will also help with the validate sbom function where we can check the file suffix to match with the known formats. Lmk what you think.

dmikusa-pivotal: Sounds reasonable. Go question for you about this. How would you achieve this? I thought you couldn't have a const slice. I know I could define the strings as const and use them in the slice, maybe that's better? Is there a common pattern for doing something like this that you've seen? Thanks

build.go
dmikusa-pivotal:
```suggestion
	sbomFiles, err := filepath.Glob(filepath.Join(layersPath, "*.sbom.*"))
```

build.go
```diff
@@ -338,6 +343,11 @@ func Build(builder Builder, options ...Option) {
 			}
 		}
 
+		// even if there is data, do not write a BOM if we have buildpack API 0.7, that will cause a lifecycle error
+		if API == "0.7" {
+			launch.BOM = nil
```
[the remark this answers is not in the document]
dmikusa-pivotal: I was initially thinking that might cause some extra noise in the output for end-users, at least as buildpacks authors are updating and working through upgrading their buildpacks.

I'll go ahead and add it. Hopefully, this code won't be around that long as we're working at v2 and we can deprecate this old BOM stuff.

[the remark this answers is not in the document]
dmikusa-pivotal: Done

layer.go
dmikusa-pivotal: I moved the strings to consts & used the consts in the array. I did a quick google search and const arrays/slices are not a thing, as far as I can tell, so I'm not sure we can do much more. I don't think using a global var would be good. I think that could potentially allow someone to modify the list, which wouldn't be desirable, IMHO.

samj1912: https://github.com/dmarkham/enumer looks like a good way to manage auto generation of all these methods.

dmikusa-pivotal: That project looks interesting, but I don't see a way to do custom transformations. If we were going from SomeConst and wanted the string representation to be `SomeConst` or `some_const` or `some-const`, it looks like we could use this. From what I see in the docs, there's no custom transformation to make `SomeConst` turn into `some.json` and vise versa, which is what we need.

It [also looks like it's generating a global var](https://github.com/dmarkham/enumer/blob/master/enumer.go#L104) which would have the same limitation, someone could potentially mess with the list.

The lists are fairly small and shouldn't grow a ton over time. It seems like we could keep them hand-written and not take on too much burden, IMHO (adding dependencies for this functionality also has its own risks). 

I'll acknowledge that there is some room for error with using the arrays to convert from enum to string, and a new type needs to be added. If you forget to update the arrays, you'll get a runtime panic. I could change that to use switch statements instead. It's more code, but when a type is added if you forget to update the conversion methods, you will end up with an unknown value instead of a runtime panic so it's a little safer. Would that be preferred?

build.go
```diff
@@ -380,3 +398,27 @@ func contains(candidates []string, s string) bool {
 
 	return false
 }
+
+func ValidateSBOMFormats(layersPath string, acceptedSBOMFormats []string) error {
```
samj1912, line 402:
```suggestion
func validateSBOMFormats(layersPath string, acceptedSBOMFormats []string) error {
```

samj1912 (approved):

build.go
samj1912: Can we make this private?

dmikusa-pivotal: Technically yes. It makes testing much harder though because it's being used in build, so you have to set up and run the entire build to test it. Given what it's doing, I wanted to make sure the test coverage was solid so I opted to leave it public.

samj1912: We could make it internal then? I would prefer to reduce our support surface area in terms of the public api to things that are strictly necessary - it makes it easier to modify/update code without breaking dependents.

samj1912: @dmikusa-pivotal couldn't find a good way either. Ended up just making it private and testing it the long way in #98. If that looks good, let's merge it.

PR #102, opened 2021-12-09
Removes the Application struct

ForestEckhardt:
- Currently the Application struct is just a wrapper for the WorkingDir
that is specified by the lifecycle. This adds an unneed layer of
abstracation and strays away from the terminology that is used in the
buildpacks spec. Instead the WorkingDir is just a string fields on both
the Build and Detect contexts that conveys the same information.

### DIFF
--- a/application.go
+++ b/application.go
@@ -16,12 +16,6 @@
 
 package libcnb
 
-// Application is the user contributed application to build.
-type Application struct {
-	// Path is the path to the application.
-	Path string
-}
-
 // Label represents an image label.
 type Label struct {
 	// Key is the key of the label.
--- a/build.go
+++ b/build.go
@@ -32,9 +32,9 @@ import (
 
 // BuildContext contains the inputs to build.
 type BuildContext struct {
-	// WorkingDir is the location of the application source code as provided by
+	// ApplicationPath is the location of the application source code as provided by
 	// the lifecycle.
-	WorkingDir string
+	ApplicationPath string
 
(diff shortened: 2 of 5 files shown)


dmikusa-pivotal:
I'd be OK with removing the Application struct, but I'm not a fan of the WorkingDir name. I understand your logic, trying to make it match the terminology that's in the spec, but IMHO it makes the API more confusing to work with.

As a buildpack developer, right now, I can say `context.Application.Path` and get the path to the application that's being built. It's very clear. I'd be OK with shortening that. Having `context.Application` or `context.ApplicationPath` just be the path as a string. Especially since there's nothing else on the Application object. Keep it simple, makes sense.

Changing to `context.WorkingDir` seems like a step too far to me. I don't feel like that is clear in terms of what it points to. As a build path author, I want the application location for all kinds of reasons. I'm thinking in terms of applications. If I'm using intellisense or the godocs, something that says "application" is going to be a big clue. WorkingDir not as much. That could be the app, it could be some other temp directory.

My $0.02.

samj1912:
I agree with @dmikusa-pivotal on the naming choice - happy if the rename is ctx.ApplicationPath - I would prefer not to call it ctx.Application since that was already a struct in the existing api and might be confusing.

ForestEckhardt:
@samj1912 @dmikusa-pivotal I am cool if y'all want the bike shed purple! I will put in a change right away.

dmikusa-pivotal (approved):

samj1912 (approved):

PR #106, opened 2022-01-05
Use appropriate bindings path for detect

### LINKED ISSUE #105: Setting `SERVICE_BINDING_ROOT` does not apply at detect
If you set an alternative location for `SERVICE_BINDING_ROOT` this only presently applies at build time. At detect, the standard `/platform/bindings` is always used.

Detect calls [NewBindingsFromPath](https://github.com/buildpacks/libcnb/blob/main/detect.go#L129-L134) which is given `/platform/bindings`. This method does not look at EnvServiceBindings, so it doesn't honor this env variable.

samj1912:
Signed-off-by: Sambhav Kothari <skothari44@bloomberg.net>

Fixes #105

### DIFF
--- a/detect.go
+++ b/detect.go
@@ -126,9 +126,8 @@ func Detect(detector Detector, options ...Option) {
 		logger.Debug(PlatformFormatter(ctx.Platform))
 	}
 
-	file = filepath.Join(ctx.Platform.Path, "bindings")
-	if ctx.Platform.Bindings, err = NewBindingsFromPath(file); err != nil {
-		config.exitHandler.Error(fmt.Errorf("unable to read platform bindings %s\n%w", file, err))
+	if ctx.Platform.Bindings, err = NewBindingsForBuild(ctx.Platform.Path); err != nil {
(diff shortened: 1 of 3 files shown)


detect.go
```diff
@@ -126,9 +126,8 @@ func Detect(detector Detector, options ...Option) {
 		logger.Debug(PlatformFormatter(ctx.Platform))
 	}
 
-	file = filepath.Join(ctx.Platform.Path, "bindings")
-	if ctx.Platform.Bindings, err = NewBindingsFromPath(file); err != nil {
-		config.exitHandler.Error(fmt.Errorf("unable to read platform bindings %s\n%w", file, err))
+	if ctx.Platform.Bindings, err = NewBindingsForBuild(ctx.Platform.Path); err != nil {
```
samj1912, line 129: Looking back at this - I don't like that we call this `NewBindingsForBuild` - we should update it to something better in v2. cc: @dmikusa-pivotal

dmikusa-pivotal: +1

dmikusa-pivotal (approved):

PR #108, opened 2022-01-10
Validation should only occur under API 0.7

### LINKED ISSUE #107: SBOM validation fails if buildpack API is 0.6
If you have a buildpack using buildpack API 0.6, libcnb will still perform SBOM validation even though that is a feature of buildpack API 0.7. This causes a problem because it expects the a list of supported SBOM media types to be present in buildpack.toml, but that is specific to API 0.7 so buildpacks with API 0.6 won't have it.

I think we need to [check the buildpack API before making this call](https://github.com/buildpacks/libcnb/blob/main/build.go#L309-L312). 

Thoughts?

dmikusa-pivotal:
- Adds API checks before validating SBOM format, should only happen with API 0.7+
- Adds a test to confirm validation does not run if the API is less than 0.7.

Resolves #107

### DIFF
--- a/build.go
+++ b/build.go
@@ -306,9 +306,11 @@ func Build(builder Builder, options ...Option) {
 		}
 	}
 
-	if err := validateSBOMFormats(ctx.Layers.Path, ctx.Buildpack.Info.SBOMFormats); err != nil {
-		config.exitHandler.Error(fmt.Errorf("unable to validate SBOM\n%w", err))
-		return
+	if API != "0.1" && API != "0.2" && API != "0.3" && API != "0.4" && API != "0.5" && API != "0.6" {
+		if err := validateSBOMFormats(ctx.Layers.Path, ctx.Buildpack.Info.SBOMFormats); err != nil {
+			config.exitHandler.Error(fmt.Errorf("unable to validate SBOM\n%w", err))
+			return
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


build.go
```diff
@@ -306,9 +306,11 @@ func Build(builder Builder, options ...Option) {
 		}
 	}
 
-	if err := validateSBOMFormats(ctx.Layers.Path, ctx.Buildpack.Info.SBOMFormats); err != nil {
-		config.exitHandler.Error(fmt.Errorf("unable to validate SBOM\n%w", err))
-		return
+	if API != "0.1" && API != "0.2" && API != "0.3" && API != "0.4" && API != "0.5" && API != "0.6" {
```
samj1912, line 309: Shall we just check if api == 0.7 for now? we can add semver parsing later.

dmikusa-pivotal: I went with this approach because it is future proof and doesn't need changing in the future. It is very ugly though.

I'm not opposed to just matching on 0.7, but I worry that'll cause something to break in the future.

+100 for better API matching. That should be a target for v2. We should also look at some possible patterns for coding multiple version support in sane ways. IMHO, having If version == xyz blocks all over is hard to maintain. There's got to be something better we can do there for v2.

samj1912 (approved):

PR #115, opened 2022-01-31
Adjustments for Lifecycle version 0.13.3

dmikusa-pivotal:
Starting with lifecycle 0.13.3, it is permitted to have both the old style label-based BOM information and the new style layer-based BOM information. If the buildpack API is 0.6 or older, label-based BOMs only are OK. If the buildpack API is 0.7, you may have both label-based BOM and layer-based BOM or just layer-based BOM. It is permitted to have just label-based BOM, however, that will generate a warning from the lifecycle.

This PR makes two changes:

1. It removes two checks that we were using to nil out the label-based BOM and WARN if it was set. This is OK now based on the lifecycle changes so we permit it.
2. It adds `BP_DISABLE_BOM_LABEL` which can be used to manually disable the label-based BOM. This is for the case where the label is too large and causes problems with K8s. This defaults to false, so label-based BOM is enabled by default. Setting it to true will result in no label-based BOM being included, even if the buildpacks write that information.

### DIFF
--- a/build.go
+++ b/build.go
@@ -122,6 +122,7 @@ type Builder interface {
 func Build(builder Builder, options ...Option) {
 	config := Config{
 		arguments:         os.Args,
+		bomLabel:          false,
 		environmentWriter: internal.EnvironmentWriter{},
 		exitHandler:       internal.NewExitHandler(),
 		tomlWriter:        internal.TOMLWriter{},
@@ -315,7 +316,7 @@ func Build(builder Builder, options ...Option) {
 
 	// Deprecated: as of Buildpack API 0.7, to be removed in a future version
 	var launchBOM, buildBOM []BOMEntry
-	if result.BOM != nil {
+	if result.BOM != nil && config.bomLabel {
 		for _, entry := range result.BOM.Entries {
 			if entry.Launch {
 				launchBOM = append(launchBOM, entry)
--- a/build.go
+++ b/build.go
@@ -345,12 +345,6 @@ func Build(builder Builder, options ...Option) {
 			}
 		}
 
-		// even if there is data, do not write a BOM if we have buildpack API 0.7, that will cause a lifecycle error
-		if API == "0.7" && len(launch.BOM) > 0 {
-			logger.Info("Warning: this buildpack is including both old and new format SBOM information, which is an invalid state. To prevent the lifecycle from failing, libcnb is discarding the old SBOM information.")
-			launch.BOM = nil
-		}
-
 		if err = config.tomlWriter.Write(file, launch); err != nil {
 			config.exitHandler.Error(fmt.Errorf("unable to write application metadata %s\n%w", file, err))
 			return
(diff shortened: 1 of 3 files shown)


build.go
```diff
@@ -313,9 +314,18 @@ func Build(builder Builder, options ...Option) {
 		}
 	}
 
+	var disableBOMLabel bool
+	if val, ok := os.LookupEnv("BP_DISABLE_BOM_LABEL"); ok {
```
samj1912, line 318: @dmikusa-pivotal I think the usage of this env var might be buildpacks/platform specific. I would prefer it if we didn't introduce it in libcnb and just relied on the buildpack to toggle this logic. This might be better placed in libpak for eg. LMK what you think.

dmikusa-pivotal: I'm not opposed to having a different trigger, but I feel pretty strongly that we should allow for the disabling of the label-based BOM entries. It's known to cause problems.

I went with an env variable because there is some precedent with `BP_DEBUG` and `BP_LOG_LEVEL` which are also in libcnb.

Other options that come to mind:

- A BuildContext property: https://github.com/buildpacks/libcnb/blob/main/build.go#L34-L56
- An Option callback: https://github.com/buildpacks/libcnb/blob/d3e6e18255ecfc1dac6cc473471d9cb2739124c4/config.go#L75

ekcasey (approved):

ekcasey:
Not strictly related to this PR but it looks like libcnb is providing [a helpers to create layer-specific SBOMs](https://github.com/buildpacks/libcnb/blob/main/layer.go#L149) but not for generic launch and build SBOMs. Seem like we should probably do both? cc @dmikusa-pivotal @samj1912

samj1912:
@ekcasey we do have those as well -> https://github.com/buildpacks/libcnb/blob/d3e6e18255ecfc1dac6cc473471d9cb2739124c4/layer.go#L216-L223

build.go
samj1912: @dmikusa-pivotal I think the difference between those log level env vars and the SBOM env var is that the log level env var influences the log lines that the build process outputs which cannot be controlled by the users of libcnb (for eg https://github.com/buildpacks/libcnb/blob/d3e6e18255ecfc1dac6cc473471d9cb2739124c4/build.go#L179) I think in this case, a buildpack using libcnb can just detect this env var and not choose to populate the bom in launch.toml/build.toml

I am concerned because I would like that env var to be set to a value so that labels are turned off by default completely and I imagine paketo might want something else. Given that we are venturing into buildpack author level opinions on configuration that is exposed to the end user(app dev) I think we should avoid introducing this in libcnb.

I might be fine with introducing this in a way that is exposed to the buildpack author and not an end user (i.e. not an env var)

dmikusa-pivotal: ✅ , I get where you're coming from.

I removed the env variable, added an option that a buildpack author can set when running `libcnb.Build`, set this to default to false (i.e. don't write the BOM label).

Let me know what you think.

build.go
```diff
@@ -122,6 +122,7 @@ type Builder interface {
 func Build(builder Builder, options ...Option) {
 	config := Config{
 		arguments:         os.Args,
+		bomLabel:          false,
```
samj1912, line 125: Should we instead just call this `allowDeprecated` ?

dmikusa-pivotal: IMHO, that's not specific enough. There could be other deprecated things to allow.

I'd be OK with `allowDeprecatedBOMLabel` and `WithDeprecatedBOMLabel`. I can make that change and resubmit in the morning if you're Ok with that.

For what it's worth, I did flag `WithBOMLabel` as deprecated, so if it's used it'll show up as deprecated in your IDE. Is that sufficient? or would you still prefer it in the variable name?

samj1912:
@dmikusa-pivotal merging this for now. Was just a nitpick on the naming side. Hopefully we are getting rid of v1 soon anyway.

samj1912 (approved):

PR #120, opened 2022-03-05
Catch up to main with v2 SBOM Changes

dmikusa-pivotal:
This PR catches up the v2 branch to the main branch in terms of SBOM support.

Here are the highlights:

- Adds support for allowing buildpack API 0.7
- Adds validation of the SBOM format based on what is in buildpack.toml

I intentionally left multiple commits on this PR so you could see which ones were pulled over from main. I ran them in order so we ended up with the same thing on v2 (except where things have diverged around build). If you want to squash these down when merging, feel free.

### DIFF
--- a/build.go
+++ b/build.go
@@ -274,7 +274,7 @@ func Build(build BuildFunc, options ...Option) {
 		}
 	}
 
-	if API != "0.1" && API != "0.2" && API != "0.3" && API != "0.4" && API != "0.5" && API != "0.6" {
+	if API != "0.5" && API != "0.6" {
 		if err := validateSBOMFormats(ctx.Layers.Path, ctx.Buildpack.Info.SBOMFormats); err != nil {
 			config.exitHandler.Error(fmt.Errorf("unable to validate SBOM\n%w", err))
 			return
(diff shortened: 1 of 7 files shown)


build.go
```diff
@@ -151,8 +151,8 @@ func Build(build BuildFunc, options ...Option) {
 	logger.Debugf("Buildpack: %+v", ctx.Buildpack)
 
 	API := strings.TrimSpace(ctx.Buildpack.API)
-	if API != "0.5" && API != "0.6" {
-		config.exitHandler.Error(errors.New("this version of libcnb is only compatible with buildpack APIs 0.5 and 0.6"))
+	if API != "0.5" && API != "0.6" && API != "0.7" {
```
samj1912, line 154: Do we want to meet libcnb v2 to just api 0.7 and greater?

samj1912 (approved):
LGTM. We should decide later on which buildpack API versions we want to support in v2.

build.go
dmikusa-pivotal: Yes, I think we should talk more about that. My $0.02, I'm torn. Part of me thinks it would be good to start fresh and remove some of the cruft from the older APIs. Part of me thinks the overhead is small and why not support more APIs?

Definitely curious to see what others think. I'll raise this up in the next BAT meeting.

PR #124, opened 2022-03-11
Make a configurable Logger interface

### LINKED ISSUE #113: libcnb v2: logging interface
Right now libcnb has the poet module for logging. It will write logging information, much of it is at the debug level (there are some things like warning messages that get written at the info level), and this information uses the format and output defined in libcnb.

If you are wrapping libcnb, like with [libpak](https://github.com/paketo-buildpacks/libpak) or like we'd like to do with packit then you end up with logging that is disjoint. Indentations don't line up nicely, colors may be off, etc...

This issue is to discuss how we can structure libcnbv2 such that logging output can be controlled in a more uniform manner.

A couple of options have been discussed in the BAT meeting, here are some notes:

1. Expose a logging facade from libcnb. Libcnb writes logs using the logging facade. Libcnb would provide a basic implementation of the facade for use in cases where only libcnb is used. In the case of extending libcnb, the extending library could provide its own implementation of the facade. The implementation would dictate formatting, indentation, color, etc.. The control over how things look would be made consistent across all buildpacks using that implementation.

    This has a couple of drawbacks. We'd have to expose our own facade, which is yet another logging API. It would also not solve the case where you have a composite buildpack made up of buildpacks that use different implementations, you could still have disjoint output there.

2. There was also some discussion about using Open Telemetry for logging. This has the advantage that there's already an API to use in buildpacks. Everyone would essentially use this API, logs would be written out in a structured format. Then based on some yet to be determined configuration, possibly set through the builder or a composite buildpack, the structured format could optionally be translated by the platform into a more human-friendly format. The format could control things like indentation, colors, etc... 

    The main complication here is that this requires changes and adoption by the platforms. We'd also have to consider backward compatibility. If running on an older platform, what happens to the output?

dmikusa-pivotal:
Presently, we have `log.Logger` which is a concrete implementation of a very basic logger. We use this in libcnb to log information. The buildpack calls only debug related methods, with one exception and that is a single call to an info method for logging a warning.

This PR proposes the following changes:

1. We have a Logger interface that defines debug methods only (more on this later). This can be implemented to control how libcnb logs.
2. We provide a basic implementation of Logger called PlainLogger that writes to stdout if `$BP_LOG_LEVEL=debug` or `$BP_DEBUG=true`. This is in the log module, so it can be bypassed if not needed.
3. We enable a Logger to be passed in through config options to the main/build/detect functions. Thus a buildpack author can provide their own implementation & control formatting.
4. We set a style rule that internally libcnb only writes debug level messages. This removes the need for Info/Warn level methods on the interface.
5. We set a style rule that internally libcnb does not write warnings. Any warning is an error. This is because the warnings that have been used are not messages to end-users, but messages to buildpack authors. Thus making it an error should force buildpack authors to deal with the problem. This should reduce user confusion as they won't see warning messages which they can do nothing about.

Resolves #113 


config.go
```diff
@@ -62,13 +66,31 @@ type ExecDWriter interface {
...
+// Logger is the interface implement by a type that wishes to write log messages generated by libcnb
+type Logger interface {
+	// Debug formats using the default formats for its operands
+	Debug(a ...interface{})
+
+	// Debugf formats according to a format specifier
+	Debugf(format string, a ...interface{})
+
+	// DebugWriter returns a writer where everything is written at this level
+	DebugWriter() io.Writer
```
ryanmoran, line 80: I can't find where this is used in the codebase. Do we need this in the interface?

dmikusa-pivotal: I don't think it's used in libcnb at the moment, so it's something we could potentially cut. I had mostly kept it as it was in the original log implementation.

The case where this tends to be useful is when running external commands. You can fetch the debug writer and pass that in as the Writer for output of the command. Looking at libcnb as a go implementation of the spec, it's probably unlikely we'd need this. There's nothing I can think of in the spec, at the moment, that would require us to run an external command.

I'll take that as one vote for removing it. I don't feel particularly strongly either way. Does anyone else feel strongly one way or another?

samj1912: Happy to get this removed as well.

detect.go
```diff
@@ -117,36 +115,36 @@ func Detect(detect DetectFunc, options ...Option) {
 	}
 
 	ctx.Platform.Path = config.arguments[1]
-	if logger.IsDebugEnabled() {
-		logger.Debug(PlatformFormatter(ctx.Platform))
+	if config.logger.IsDebugEnabled() {
```
samj1912, line 118: I am not sure why the existing code has double the checks for debug enablement.

config.logger.Debug should be able to handle this on its own without having a guarding if condition. I would vote that we remove this and just call config.logger.Debug (which can internally decide whether or not to log this based on whether debug is enabled).

I would also vote we get rid of `IsDebugEnabled` from the public logger interface.

dmikusa-pivotal: Classically log APIs have an `IsDebugEnabled` style method so that you can avoid debug logging when it's expensive. Not so much the performance of the log operation itself, but if you want to log some output that is expensive to generate. Perhaps for debugging, I want to log all of the files in the application directory, this could be a lot, and walking the file system isn't super speedy. That would be expensive to generate, so I don't want to incur that cost unless debug is actually enabled. Being able to check `IsDebugEnabled` allows me to skip even generating that data. Without it, I'd end up generating the data, calling `Debug` and throwing it away.

Also, less important because it's an implementation detail, but the prior implementation would segfault if you call `logger.Debug` but do not actually have debug logging enabled. Because the call to `logger.Debug` just writes directly to the "debug" writer, and that writer is nil if you don't have debug enabled. I will check and see if I fixed this, I don't think so, as I'm just remembering this fun detail.

I'll fix it so that you don't have to call `IsDebugEnabled` to not crash, but I do think we should keep `IsDebugEnabled` as it does provide an author some additional flexibility. 

The individual places `IsDebugEnabled` is being used could be evaluated though and cleaned up if the operation it's guarding isn't actually expensive to calculate. That would make for cleaner code.

config.go
dmikusa-pivotal: Cool, I'll pull it out.

detect.go
samj1912: But that still begs the question if that is something that should be a part of the logger interface or just a general flag that is exposed as a part of the build context?

samj1912: Thinking about it again, I guess if we are debugging things, the primary way to interact with the user is through logs. So I agree that it is fine to keep it attached to the logging interface.

samj1912:
Also related https://github.com/buildpacks/libcnb/issues/130

@dmikusa-pivotal do we just want to standardize on this interface?

detect.go
dmikusa-pivotal: The flag to turn it on/off is `BP_LOG_LEVEL`, setting it to `debug`, so that is tied directly to logging. 

Possibly we want to discuss the name of this flag? We do still support `BP_DEBUG` as well. That is the older flag though and I was planning to ask if we could remove it in v2 and just go with `BP_LOG_LEVEL`. Maybe we double down on `BP_DEBUG` and get rid of `BP_LOG_LEVEL` so it's not tied directly to logging? We also only have one log level, so it's really only on or off. `BP_DEBUG` fits better in that way too.

Regardless, I still think it makes sense to keep that state on the logger. I can't think of a case where I'd be doing something for the purpose of debugging and not have it generate some amount of logging. If I'm calling `logger.Debug` then it is convenient to be able to check the same logger if debug is enabled, rather than look somewhere else. I may be in a method that doesn't have access to the other state, like context, but I'll have the logger to be able to log stuff.

build.go
```diff
@@ -123,32 +122,31 @@ func Build(build BuildFunc, options ...Option) {
 		ok   bool
 	)
 	ctx := BuildContext{}
-	logger := log.New(os.Stdout)
```
samj1912, line 126: Do we want to set a default logger in config at L103?

dmikusa-pivotal: I avoided this because I didn't want anything to consume libcnb's `log` package by default. If we add a default, then either config or main references the `log` package. 

I know that Go can do some smart things around stripping out code that is unused to reduce binary size. I don't know Go well enough to know exactly what it can do though. My fear was that by including a default, we'd be forcing Go to always compile in the code for this default log implementation (what in the `log` package), even if a buildpack author provides their own logging implementation. That fear may be unfounded, maybe Go is smart enough to still strip it out but I didn't know how to test it either, so I just left it out.

Right now, the logger gets passed into the `func Main(...)` method as an `Option` and then set on the `Config` object, but that makes it optional. Maybe the way to go here is to make it a required argument to `func Main(...)` so that it's never nil. 

Thoughts?

samj1912: I think just having some check to make sure that the default is not nil would be nice. I would personally prefer if we had a sane default of "plainlogger" but I can understand the reasoning behind not wanting to import that.

dmikusa-pivotal: I can certainly add a nil check. It's not the nicest experience though because I basically have to fail and say there's no logger.

Tbh, I don't have really strong feelings about the whole import situation. I had heard that come up before as a concern and so I thought I would try to keep that into consideration with the design. I have a gut feeling that it's probably a premature optimization given the size of the log package, which is very small, but again, I don't really know how to measure/test/prove that either besides just compiling with and w/out the import and checking the difference.

If you feel like the ergonomics/usage of the API is better by having a default, I'd be OK with that option too.

dmikusa-pivotal: I was playing around with [this tool](https://github.com/knz/go-binsize-viz) for examining the size of a buildpack binary. I looked at a couple of the main binaries for Paketo buildpacks. 

Here are some stats from one:

```
19% runtime/
16% github.com/
15% crypto/
13% net/
8.4% vendor/golang.org/x
```

Then a bunch of stuff in the 3.5% and lower range. Things like `encoding/`, `regexp/`, `math/`, `time/`, `os/`, etc...

Within `github.com/`

```
3.9% paketo-buildpacks/
3.9% pelletier/go-toml
3.5% BurntSushi/toml
1.2% or 40k buildpacks/libcnb
1.2% h2non/filetype
```

Within `buildpacks/libcnb` 0.044% or 1.5k is poet.

I guess based on that, it's probably not really critical to separate these packages. The additional overhead of a few small classes is just going to be insignificant. I'll go ahead and make that change.

dmikusa-pivotal: Default added.

samj1912 (approved):

PR #131, opened 2022-04-04
Add support for buildpack api 0.8

### LINKED ISSUE #128: Support Buildpacks API 0.8: switch from build/detect args to env variables
- When using buildpack API 0.8 or greater, the platform directory and build plan path are provided as environment variables to buildpack bin/detect, in addition to the deprecated positional arguments (https://github.com/buildpacks/lifecycle/pull/818)
- When using buildpack API 0.8 or greater, the platform directory, buildpack plan path, and buildpack layers directory are provided as environment variables to buildpack bin/build, in addition to the deprecated positional arguments (https://github.com/buildpacks/lifecycle/pull/818)

samj1912:
Resolves #128


build.go
```diff
@@ -191,11 +186,36 @@ func Build(builder Builder, options ...Option) {
...
+		}
+		buildpackPlanPath, ok = os.LookupEnv("CNB_BP_PLAN_PATH")
+		if !ok {
+			config.exitHandler.Error(fmt.Errorf("expected CNB_BP_PLAN_PATH to be set"))
+			return
+		}
+	}
 
-	ctx.Platform.Path = config.arguments[2]
+	logger.Debugf("Layers: %+v", ctx.Layers)
```
dmikusa-pivotal, line 218:
```suggestion
	logger.Debugf("Layers Directory: %+v", ctx.Layers)
```

dmikusa-pivotal (approved):
Looks good to me.

I'm not 100% sure it's necessary, but my only observation is that in the detect & build tests, we're not testing the case where both args & env variables are set.

I believe that the code is going to take the env variables if both are set. It's an invalid state, so it would seem reasonable if we errored if both are set. I don't know if the platform/lifecycle would allow this (I didn't test), but my line of thinking is that it might be possible for a user to set one of these new env variables on an older platform/lifecycle since the older platform/lifecycle is unaware of them altering the behavior. In this particular case, erroring out would be beneficial.

samj1912:
> Looks good to me.
> 
> I'm not 100% sure it's necessary, but my only observation is that in the detect & build tests, we're not testing the case where both args & env variables are set.
> 
> I believe that the code is going to take the env variables if both are set. It's an invalid state, so it would seem reasonable if we errored if both are set. I don't know if the platform/lifecycle would allow this (I didn't test), but my line of thinking is that it might be possible for a user to set one of these new env variables on an older platform/lifecycle since the older platform/lifecycle is unaware of them altering the behavior. In this particular case, erroring out would be beneficial.

Even if the user sets this on an older version of the platform/lifecycle, we only source from the env var if the buildpack api >= 0.8.

build.go
samj1912: Is it okay if we keep it to the previous statement since that is what libcnb was logging before?

dmikusa-pivotal: Oh, I thought that had changed in the diff. Yes, not a huge issue. Just thought that would be a little clearer.

PR #133, opened 2022-04-05
Add support for working-directory in launch.toml (API>=0.8)

### LINKED ISSUE #127: Support Buildpacks API 0.8: working directory per process
When using buildpack API 0.8 or greater, buildpacks can specify the working directory for processes in launch.toml (https://github.com/buildpacks/lifecycle/pull/792)

gcemaj:
Add support for working directory setting in launch.toml

If API is not 0.8 then log that this setting will be ignored, and reset working directory to be blank. Otherwise ensure that the setting is propagated.

Closes #127


build_test.go
```diff
@@ -549,6 +549,152 @@ version = "1.1.1"
...
+					Default:          true,
+					WorkingDirectory: "/my/directory/",
+				},
+			},
+			Slices: []libcnb.Slice{
+				{
+					Paths: []string{"test-path"},
+				},
+			},
+		}, nil)
```
samj1912, line 659:
```suggestion
			Processes: []libcnb.Process{
				{
					Type:             "test-type",
					Command:          "test-command-in-dir",
					Default:          true,
					WorkingDirectory: "/my/directory/",
				},
			},
		}, nil)
```

samj1912: Could we make appropriate changes in the rest of the places to reduce the testing code to be limited to the process working dir bits?

build.go
```diff
@@ -346,6 +346,15 @@ func Build(builder Builder, options ...Option) {
 			}
 		}
 
+		if API != "0.8" {
+			for i, process := range launch.Processes {
+				if process.WorkingDirectory != "" {
+					logger.Infof("WARNING: Launch layer is setting working-directory=%s, but that is not supported until API version 0.8. This setting will be ignored.", process.WorkingDirectory)
```
dmikusa-pivotal, line 352: Should this be an error? If a parameter is being set but is not supported, then that could cause some unexpected behaviors. 

I'm also not sure that this is something an end-user would understand. It's probably directed at buildpack authors & a failure would more quickly get the author's attention.

gcemaj: i was following the format of the process.Default setting. if you feel this should be an error i can change it.

gcemaj: also, making it an error would make this backwards incompatible right?

dmikusa-pivotal: Yes, you're right about the backward compatible note. I had this logic backward in my head. I was thinking this only applied to 0.8+. 

I still don't totally like it though. A user can't set the working directory, the buildpack would set that so the message isn't really helpful and could be confusing to a user. 

It also looks like the lifecycle is going to [emit a very similar warning](https://github.com/buildpacks/lifecycle/pull/792/files#diff-e348e4b0ed5d889979d2fb076558d56656b133c84892137fdd110002e8948ec8R323-R325).

Given that, I think we should just leave it out and let lifecycle handle it. @samj1912 what do you think?

samj1912: I am not sure why it would be backwards incompatible to throw an error if this field is set for buildpacks api `<0.8`. They shouldn't be setting this field anyway. I think in general we need to figure out a good way to gate new features in libcnb based on buildpack api versions without spitting out the same warning as the lifecycle or duplicating validation code. For now given that the default attribute follows a similar strategy (warn and set to empty) I think it's fine to keep it this way in libcnb v1. We can figure out some stronger validation in v2 that can provide feedback to buildpack authors early on.

dmikusa-pivotal: >I am not sure why it would be backwards incompatible to throw an error if this field is set for buildpacks api <0.8. They shouldn't be setting this field anyway.

I was thinking that if we output an error where an error wasn't possible before, you could potentially break buildpacks that exist. But you're right that for existing buildpacks, they should not be writing to this field since it didn't exist previously.

dmikusa-pivotal: I'm OK with leaving this in v1 and changing the behavior in v2.

samj1912 (approved):

PR #134, opened 2022-04-05
Add semver checks

gcemaj:
Use semver for version checks instead of strings

Closes: #132

PR #140, opened 2022-05-18
Adds SBOM Formatter

ForestEckhardt:
This PR adds an SBOM Formatter similar to the one found in the (`packit/v2` implementation)[https://github.com/paketo-buildpacks/packit/blob/v2/sbom_formatter.go]. I believe that the current `libcnb` implementation is far too sparse leaving a lot of work up to either buildpack authors or library authors to decide the best way of handling the process of writing SBOMs. I believe that this interface is powerful because it allows for authors to still implement custom SBOMs however it abstracts away much of the boiler plate when it comes to having to actually manage and write the SBOM files. With this interface an author only needs to focus on writing a formatter that conforms to the `io.Reader` interface everything else will be handled for them by the library.

I would love this PR to be a conversation starter for a more unified SBOM experience as I see this as being one of the last barriers for `packit` being able to be built on top of `libcnb`.


ForestEckhardt:
> 1. I think this needs to be rebased. There appear to be unrelated file changes happening from [Removes usage of ioutil #139](https://github.com/buildpacks/libcnb/pull/139), which makes it a little hard to follow this PR.

Yeah that is my bad got ahead of myself. Rebased and hopefully should be easier to follow not.

> 2. I don't really want to remove the existing SBOM functionality like the format consts and the methods for getting paths to the SBOM locations. Similarly, I still think there's value in the validation of SBOM types. I'm not opposed to adding new functionality, but I think we should build on what's there.

So would you just like to support both methods of setting the SBOM. Support a SBOM formatter but also allow authors to just fetch a SBOM path and then arbitrarily write a file there? I am a little worried that might muddy the API so I am curious what you feel the advantage of keeping the old API around.

I am willing to go in and add back SBOM type validation but I would like to hear the argument for keeping the path fetching method. 

> 3. Do you have any examples of how the SBOMFormat/SBOMFormatter is being used in actual buildpacks? I'm not sure I'm getting a good understanding from the tests. I'm curious how it can be used in a buildpack.

This is the formatter that is used by most `packit/v2` buildpacks that have layer SBOM support. An example of where we use this is [go-build in `build.go`](https://github.com/paketo-buildpacks/go-build/blob/main/build.go). Here we use a custom `packit` SBOM formatter to scan the layer that contains all of our built binaries and then with that one scan we are able to generate multiple SBOM formats. We do this so that we only need to scan the app with Syft one and then using the data from that one scan we can generate all of the SBOMs that we need.

build.go
```diff
@@ -150,8 +152,14 @@ func Build(build BuildFunc, options ...Option) {
...
-	API := strings.TrimSpace(ctx.Buildpack.API)
-	if API != "0.5" && API != "0.6" && API != "0.7" {
+	apiV05, _ := semver.NewVersion("0.5")
+	apiV06, _ := semver.NewVersion("0.6")
+	apiVersion, err := semver.NewVersion(ctx.Buildpack.API)
+	if err != nil {
+		config.exitHandler.Error(err)
+		return
+	}
+	if apiVersion.LessThan(apiV05) {
```
dmikusa-pivotal, line 162: I think this is something that didn't get ported over from v1.

https://github.com/buildpacks/libcnb/pull/134

That PR needs to be applied to v2 as well, before we can merge this. Otherwise, I think it's going to create merge conflicts.

dmikusa-pivotal:
>So would you just like to support both methods of setting the SBOM. Support a SBOM formatter but also allow authors to just fetch a SBOM path and then arbitrarily write a file there? I am a little worried that might muddy the API so I am curious what you feel the advantage of keeping the old API around.

To me, the reason I think we need to keep the existing API around is that it's not *just* for writing SBOMs. It's for anything where you might want to get the SBOM path. I definitely use it in testing. Ex: test cases to exercise conditional logic to write or not write the SBOM files. It may be less relevant with the introduction of the formatters, but the point is that having convenient ways to get the SBOM paths can be useful for other reasons as well. Since libcnb wraps the API, it seems natural to provide convenient functionality like this.

I'm pretty sure that we had originally discussed this style of API as well, where the actual writes are abstracted away. I'm trying to remember why we opted not to do that. I can't exactly remember though, sorry. Not sure if @samj1912 recalls. Definitely curious to hear his thoughts on this.

build.go
ForestEckhardt: Just added this construct but none of the other 0.8 API changes. Let me know if this will suffice.

sbom_formatter.go
```diff
@@ -0,0 +1,27 @@
...
+type SBOMFormat struct {
+	Extension string
+	Content   io.Reader
+}
+
+// SBOMFormatter defines the interface for types capable of generating
+// formatted SBoMs.
+type SBOMFormatter interface {
+	Formats() []SBOMFormat
+}
```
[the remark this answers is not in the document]
ForestEckhardt: Added comments and mocks for both of the new interfaces.

sbom_formatter.go
```diff
@@ -0,0 +1,27 @@
...
+
+import "io"
+
+// SBOMFormat represents the mapping of a formatted SBOM content to its file
+// extension on the filesystem.
+type SBOMFormat struct {
+	Extension string
+	Content   io.Reader
+}
+
```
dmikusa-pivotal, line 11:
```suggestion

//go:generate mockery --name=SBOMFormatter --case=underscore

```

We typically add this comment. That should make it generate the mock.

sbom_formatter.go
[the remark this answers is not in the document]
dmikusa-pivotal: I don't think it picked up this interface, you need the go comment entry (see my other comment). Then when you run `go generate ./...` it should add a file under `mocks` for sbom_formatter.go.

build.go
dmikusa-pivotal: Give me a min, I'll port #134 over. I think if you only add this bit, it'll make merging 134 more difficult. If we merge that over first, then rebase your PR, I think these diffs should just disappear.

sbom_formatter.go
ForestEckhardt: Yeah that is my bad. I just forgot to write the file one second.

sbom_formatter.go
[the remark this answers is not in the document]
ForestEckhardt: Alright I think I got it in there this time.

build.go
dmikusa-pivotal: Ok, I have #142 that ports the recent changes from main into version-2. My suggestion would be that we merge that first and rebase this PR off that. It does have a few changes to build, but I think there will be fewer conflicts if we go in this order.

sbom_formatter.go
dmikusa-pivotal: Looks good, thanks.

build.go
ForestEckhardt: Sounds good to me.

ForestEckhardt:
@dmikusa-pivotal Alright everything should be merged to be more understandable!

samj1912:
@ForestEckhardt there are also some open questions about the way that packit implements SBOMs https://paketobuildpacks.slack.com/archives/C011S6EL49L/p1653411685386879

should we table this PR until at the very least the relevant parties can meet and discuss the API around the current packit implementation and its potential pros/cons?

ForestEckhardt:
@samj1912 Is there any update on the discussions around this PR?

dmikusa:
This PR is out of date, so I'm going to close it. 

I started a conversation on [Slack](https://cloud-native.slack.com/archives/C0333LG1E68/p1717694738639239) so we can discuss this more and what this might look like in the library.

PR #141, opened 2022-05-18
Adds layer.Reset function

ForestEckhardt:
The `layer.Reset()` function is some functionality that exists in (`packit/v2` library)[https://github.com/paketo-buildpacks/packit/blob/7b9a5b045761a8d224580cdb904e5464fa0e8f8b/layer.go#L95-L117]. It allows a buildpack author to ensure that there is no data left over from a cached layer. I felt that it would be appropriate to add this functionality over here as well for authors who want to guarantee that their layers are empty.

### DIFF
--- a/layer.go
+++ b/layer.go
@@ -156,6 +156,8 @@ func (l Layer) Reset() (Layer, error) {
 	l.SharedEnvironment = Environment{}
 	l.BuildEnvironment = Environment{}
 	l.LaunchEnvironment = Environment{}
+	l.Profile = Profile{}
+	l.Exec = Exec{Path: filepath.Join(l.Path, "exec.d")}
 	l.Metadata = nil
 
 	err := os.RemoveAll(l.Path)
--- a/layer.go
+++ b/layer.go
@@ -157,7 +157,6 @@ func (l Layer) Reset() (Layer, error) {
 	l.BuildEnvironment = Environment{}
 	l.LaunchEnvironment = Environment{}
 	l.Profile = Profile{}
-	l.Exec = Exec{Path: filepath.Join(l.Path, "exec.d")}
 	l.Metadata = nil
 
 	err := os.RemoveAll(l.Path)
--- a/layer.go
+++ b/layer.go
@@ -146,6 +146,31 @@ type Layer struct {
 	Exec Exec `toml:"-"`
 }
 
+func (l Layer) Reset() (Layer, error) {
+	l.LayerTypes = LayerTypes{
+		Build:  false,
+		Launch: false,
+		Cache:  false,
+	}
+
+	l.SharedEnvironment = Environment{}
+	l.BuildEnvironment = Environment{}
+	l.LaunchEnvironment = Environment{}
+	l.Metadata = nil
+
+	err := os.RemoveAll(l.Path)
+	if err != nil {
+		return Layer{}, fmt.Errorf("error could not remove file: %s", err)
+	}
+
+	err = os.MkdirAll(l.Path, os.ModePerm)
+	if err != nil {
+		return Layer{}, fmt.Errorf("error could not create directory: %s", err)
+	}
+
+	return l, nil
+}
+
 // SBOMPath returns the path to the layer specific SBOM File
 func (l Layer) SBOMPath(bt SBOMFormat) string {
 	return filepath.Join(filepath.Dir(l.Path), fmt.Sprintf("%s.sbom.%s", l.Name, bt))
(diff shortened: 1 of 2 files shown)


layer.go
```diff
@@ -146,6 +146,31 @@ type Layer struct {
...
+		Build:  false,
+		Launch: false,
+		Cache:  false,
+	}
+
+	l.SharedEnvironment = Environment{}
+	l.BuildEnvironment = Environment{}
+	l.LaunchEnvironment = Environment{}
+	l.Metadata = nil
+
```
samj1912, line 160: Do we also want to reset Profile?

dmikusa-pivotal:
Generally 👍 on this.

Two questions. 

1. Do you think it would be worthwhile to have either an argument or a similar function that only clears out the contents of the layer? i.e. just remove/recreate the layer directory.

I'd have to dig deeper to confirm, but I think there are cases in libpak where we are doing something similar.

2. I'm wondering about what the consumption of `layer.Reset()` looks like in packit. What situations do you find yourself using this? Do you generally have a layer you're actively working with and then want to reset it? or are you creating a new layer & resetting it? 

    Right now libcnb has `layers.Layer(..)` which creates a new layer that creates a layer and if it exists loads metadata. I could perhaps see expanding on that to allow creating an empty layer, or creating a layer and wiping any existing metadata/files.

    My initial thought was what is the difference between creating a new layer and resetting a layer, and I don't think much except the context in which that's being used by library consumers. If buildpack authors are having to create a new layer and reset it immediately, that's not great. If on the other hand the pattern is to create a layer, use it a bit, then perhaps reset it, I could see the reset method being handy. Or perhaps we could have both? Thoughts?

ForestEckhardt:
So the general workflow that we have set up in most of our buildpacks that supply dependencies is that we will load a layer with `layers.Get()` we will then look at the metadata and see if we should reuse the layer. If the layer needs to be rebuilt because the cached layer does not line up we will then call `layer.Reset()` to clear the layer and make sure it is clean for the new dependency. A good example of this is [go-dist `build.go`](https://github.com/paketo-buildpacks/go-dist/blob/main/build.go). 

There are cases where we just want to load a layer and use it without doing any cache evaluation and then just add information to it. Typically we would see that type of layer for something like a build cache or a module cache. A good example is in [go-build `build.go`](https://github.com/paketo-buildpacks/go-build/blob/main/build.go). Here we create a `goCacheLayer` that we use to store all of the `go build` cache which we use to make rebuilds faster. 

The real purpose for `layer.Reset()` is to ensure that your rebuilds are not dirtied by previous builds when they shouldn't be. It makes the author have to be more intentional about keeping their image clean and gives them the ability to ensure cleanliness.

dmikusa-pivotal:
Gotcha. Makes sense. I think this is probably a new pattern in libcnb. Previously, with the layer contributors wrapped handled some of these decisions. Thanks

dmikusa-pivotal (approved):

layer.go
ForestEckhardt: Added both Profile and Exec to the reset logic. Exec might not be necessary but I figured I would add it for consistency sake.

samj1912:
@ForestEckhardt I have a couple of qs. I think the current API in libcnb is a bit of a mistake where we allow the execd field to be public writeable. It should ideally be read only. Given the above, we should not be changing that during reset and ideally also fixing this in v2 in a subsequent PR.

Could you make it so that reset doesn't change the execd path?

samj1912:
I am curious what y'all think about an API like this -> https://samj1912.github.io/python-libcnb/api/#libcnb._layers.Layer.load

Do we also need an equivalent load function here as well?

samj1912 (approved):

ForestEckhardt:
> I am curious what y'all think about an API like this -> https://samj1912.github.io/python-libcnb/api/#libcnb._layers.Layer.load
> 
> Do we also need an equivalent load function in libcnb as well? (Maybe in a subsequent PR?)

I think that this is what `layers.Layer` does. It does however fail to load the previous environment variables which is something we should address.

samj1912:
> > I am curious what y'all think about an API like this -> [samj1912.github.io/python-libcnb/api/#libcnb._layers.Layer.load](https://samj1912.github.io/python-libcnb/api/#libcnb._layers.Layer.load)
> > Do we also need an equivalent load function in libcnb as well? (Maybe in a subsequent PR?)
> 
> I think that this is what `layers.Layer` does. It does however fail to load the previous environment variables which is something we should address.

Do we also want to load profile apart from env vars?

PR #145, opened 2022-05-27
Removes TODOs around CNB_BUILDPACK_DIR

### LINKED ISSUE #52: Deprecate code from old Buildpack API versions
Currently libcnb only supports api version 0.6 but it still has code that supports paths specific to old api versions (marked by various TODOs in the codebase). We should clean up this code if we don't plan on supporting api versions older than 0.6 moving forward.

dmikusa-pivotal:
The lifecycle has had CNB_BUILDPACK_DIR support since 0.9.0. It seems safe to remove this fallback code & just fail if the env variable is not present.

Resolves #52. There are no more TODO notes in the code base.


layer_test.go
```diff
@@ -17,6 +17,7 @@
 package libcnb_test
 
 import (
+	"io/ioutil"
```
samj1912, line 20: Was this needed?

samj1912 (approved):
LGTM, minor comment.

layer_test.go
dmikusa-pivotal: Yes, Idk how that happened, but the test was failing locally for me.

samj1912 (approved):

PR #146, opened 2022-05-27
Move formatters out of main package

dmikusa:
Thoughts on this one?

-----

The primary purpose of this PR is to move the formatters that were only used for debug logging out of the main package.

Doing this has a few other side effects that are helpful as well:

- It moves the Logger interface out of the main package into Log. It didn't make a lot of sense that this interface was not in the log package. It was originally put there because that would not require libcnb to use the log package, but we are doing this anyway so we can default to the plain logger.
- It adds a formatter interface so one could customize the output of the directory contents. This is important if you are adding customization to the output format on top of libcnb, i.e. making your own logger.
- DirectoryContentsWriter is more generic now. It doesn't need customized for every directory to log.
- DirectoryContentsWriter now writes directly to a writer, where it previously build up a large string in memory. This could use an unbounded amount of memory given that we log contents of user controlled locations like the application directory.
- Errors are handled in a more sane way. They are still swallowed, but they bubble up to the appropriate location and are captured.


detect.go
```diff
@@ -97,7 +104,12 @@ func Detect(detect DetectFunc, options ...Option) {
 	}
 
 	if config.logger.IsDebugEnabled() {
-		config.logger.Debug(BuildpackPathFormatter(ctx.Buildpack.Path))
+		if internal.NewDirectoryContentsWriter(
```
samj1912, line 107: I think we are missing an err assignment here

detect.go
```diff
@@ -139,7 +151,12 @@ func Detect(detect DetectFunc, options ...Option) {
 	}
 
 	if config.logger.IsDebugEnabled() {
-		config.logger.Debug(PlatformFormatter(ctx.Platform))
+		if internal.NewDirectoryContentsWriter(
```
samj1912, line 154: Same here

internal/directory_contents.go
```diff
@@ -18,36 +18,50 @@ package internal
...
+	path   string
+	format log.DirectoryContentFormatter
+	writer io.Writer
 }
 
-// Get returns the names of all files within a directory
-func (d DirectoryContents) Get() ([]string, error) {
-	var contents []string
+// NewDirectoryContentsWriter returns a new DirectoryContentsWriter initialized and ready to be used
+func NewDirectoryContentsWriter(path string, format log.DirectoryContentFormatter, writer io.Writer) DirectoryContentsWriter {
```
samj1912, line 36: I'm curious what your thoughts are about New* functions. Typically they make it really hard to extend the API if we add new fields to the struct.

log/formatter.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+//go:generate mockery --name DirectoryContentFormatter --case=underscore
+
+// DirectoryContentFormatter allows customization of the directory output
+type DirectoryContentFormatter interface {
```
samj1912, line 28: This is an interesting API. Could we also add some examples of how these fields affect the logged output? I'm having a hard time picturing how RootPath for example will be used by an end user without them having to dive into logging internals.

detect.go
dmikusa-pivotal: Yes, good catch. I'll fix.

detect.go
dmikusa-pivotal: I'll fix this one too.

internal/directory_contents.go
dmikusa-pivotal: On one hand, it's an established pattern in Go so I try to follow it.

I do agree that it creates challenges though. It gets unwieldy if you have more than a few arguments, and you're right it's tied very closely to the struct properties so if you change one you have to change the other. It definitely makes it harder to evolve the code without breaking the API.

What I've done in some of our Paketo code is:
    - if we add new fields, we end up adding a new New* method, and deprecating the old method. It's still there, but not breaking.
    - In some cases, we use the Option pattern, where you pass in a variable number of Options that then get applied against the struct

The Option pattern is nice with large structs that have lots of properties or to add more flexibility. The downside is that enforcing required properties then becomes a runtime check as opposed to a requirement through the function arguments.

In this particular case, I added the New method because I wanted to make the struct properties all private (I know you've mentioned wanting to do this more). This would seem to necessitate a constructor, so the struct can be created outside of its package. I'm also OK in this case as I don't think the struct properties will really need to change here. We're already passing in pretty much all the info we can for use in by the formatter.

samj1912:
@dmikusa - I made some minor changes so that the default formatter (plain formatter) is internal only since it takes no args. I changed the interface for the content writer a bit to reuse the formatter/debug writer and had the write method take in both the title and the path.

samj1912 (approved):

PR #147, opened 2022-05-27
Modifies bindings API to improve consistency

### LINKED ISSUE #73: Permission failure for platform does not get logged
We encountered an issue where a user had bindings for a build, but the binding due to some issues with Docker on MacOS had permission issues. Inside the container, the bindings were not readable.

Unfortunately, this didn't seem to cause the buildpack to exit or log anything meaningful. Only when we turned on debug logging did we see the following:

```
Layers: {Path:/layers/paketo-buildpacks_ca-certificates}
Platform contents: error walking path /platform
readdirent /platform/bindings: operation not permitted
Platform Bindings: []
Platform Environment: map[BP_DEBUG:true]
Persistent Metadata: map[]
Buildpack Plan: {Entries:[{Name:ca-certificates-helper Metadata:map[]}]}
Stack: io.buildpacks.stacks.bionic
```

which was enough to tip us off that there were problems.

It seems like this was just a side effect of logging and that the error here is not critical. It is caused by the [PlatformFormatter](https://github.com/buildpacks/libcnb/blob/main/formatter.go#L54-L56) when it tries to print the contents. That triggers a walk of the directory, which fails.

That would only happen with debug logging enabled and I think that would be expected, and no reason to fail the build.

That said, the build should fail when we're [reading the bindings](https://github.com/buildpacks/libcnb/blob/c07e39f24859ebb30db0d26ed2c555899135dfc0/platform.go#L192). From looking at the code, it appears as it should, but in the output, there is no error logged.

Here's the output from what was captured:

```
===> BUILDING
Starting build
Running build for buildpack paketo-buildpacks/ca-certificates@2.3.2
Looking up buildpack
Finding plan
Running build for buildpack Paketo CA Certificates Buildpack 2.3.2
Creating plan directory
Preparing paths
Running build command

Paketo CA Certificates Buildpack 2.3.2
  https://github.com/paketo-buildpacks/ca-certificates
  Launch Helper: Contributing to layer
    Creating /layers/paketo-buildpacks_ca-certificates/helper/exec.d/ca-certificates-helper
Processing layers
Updating environment
Reading output files
Updating buildpack processes
Updating process list
Finished running build for buildpack paketo-buildpacks/ca-certificates@2.3.2
```

IMHO, we should fail if we cannot read the bindings unless the failure is caused because the bindings don't exist. A permissions issue should fail so a user can be alerted and fix the issue.

dmikusa-pivotal:
- Removes `NewBindingsForLaunch`
- Renames `NewBindingsForBuild` to `NewBindings`
- `NewBindings` is the recommended entrypoint for loading bindings across build/detect/launch
- `NewBindingsFromPath` is being left public for historical reasons and to provide a bit of extra flexiblity
- build.go and detect.go were switch to use `NewBindings`
- Uses `os.ReadDir` instead of `filepath.Glob`, the latter will swallow I/O errors which can make it hard to debug binding issues

Resolves #73.


platform.go
```diff
@@ -147,14 +145,18 @@ func NewBindingsFromPath(path string) (Bindings, error) {
...
-func NewBindingsForBuild(platformDir string) (Bindings, error) {
+// NewBindings creates a new bindings from all the bindings at the path defined by $SERVICE_BINDING_ROOT.
+// If that isn't defined, bindings are read from <platform>/bindings.
+// If that isn't defined, the specified platform path will be used
+func NewBindings(platformDir string) (Bindings, error) {
 	if path, ok := os.LookupEnv(EnvServiceBindings); ok {
 		return NewBindingsFromPath(path)
 	}
 
+	if path, ok := os.LookupEnv("CNB_PLATFORM_DIR"); ok {
```
samj1912, line 156: Should we make this env var a public const as well?

dmikusa-pivotal: Yes, I can definitely do that.

platform.go
```diff
@@ -40,6 +40,27 @@ const (
 	// See the Service Binding Specification for Kubernetes for more details - https://k8s-service-bindings.github.io/spec/
 	EnvServiceBindings = "SERVICE_BINDING_ROOT"
 
+	// EnvBuildpackDirectory is the name of the environment variable that contains the path to the buildpack
```
samj1912, line 43: These other constants should probably be in different files?

dmikusa-pivotal: I honestly wasn't sure what makes sense (kind of paralyzed by choice, lol).

To me build.go/detect.go are where they are used most, but perhaps you could make a case they go in buildpack.go since they are part of the buildpack spec, or possibly we should have a const.go (or something like that) where all the consts in `libcnb` package go (I think I've seen that before in go projects).

In the end, I put them here to be colocated with service binding root since I followed the pattern `Env*` for the name of the const. All those locations end up in `libcnb` package, so it didn't seem to make a huge difference from the consumer side.

Do you have a preference where? I don't, so if you want me to move them, I'd be happy to.

samj1912: I am going to approve and merge this. We can move the variables around in a different PR.

samj1912 (approved):

PR #148, opened 2022-05-31
Finish removing shell-specific logic (RFC #0093)

### LINKED ISSUE #70: [RFC #0093] Remove Shell-Specific Logic
This issue have been automatically created from pull request buildpacks/rfcs#168.

dmikusa:
This PR submits the second half of RFC #0093, which is removing the Direct field from the Process struct and also switching the Command field from a string to a list of strings.

This is a breaking change as it removes a field from the struct and changes an existing field, but given this change is only going into v2+ that seems to be the correct path.

Resolves #70.


samj1912:
@dmikusa-pivotal we should probably block this PR on the lifecycle release which removes shell support.

dmikusa-pivotal:
Oh, ok. I thought that had happened already. Do you know the timing of that release? I agree we should wait, but that's going to block us from being able to cut a 2.0 release. There are definitely breaking changes required here.

Maybe we could do a milestone release in the meantime? Does Go have a mechanism for that? Or would one typically just use a specific release commit hash? Just trying to think of a way we can have something that's semi-finished for folks to start playing with, I know I would like to try getting libpakv2 ready to go soon.

samj1912:
I don't think we need to block 2.0 on this issue. I'm happy to cut a release with just BP API 0.8

samj1912:
Marking this as draft so that it is not accidentally merged till the lifecycle release.

natalieparellano:
Now that lifecycle 0.15.0 is out with Buildpack API 0.9 support, could this be un-drafted?

samj1912 (approved):

dmikusa:
OK, all the tests pass and I merged this in. @samj1912 Should we cut a release?

PR #160, opened 2022-07-14
Ensure directories are created when process-specific env variables are used

### LINKED ISSUE #158: Process type specific environment variables break EnvironmentWriter
The EnvironmentWriter is [not creating sub-directories for process-specific environment variables](https://github.com/buildpacks/libcnb/blob/release-1.x/internal/environment_writer.go#L40-L41).

This results in an error when you try to write a process specific environment variable:

```
unable to write layer env.launch /layers/paketo-buildpacks_health-checker/thc/env.launch
unable to write file /layers/paketo-buildpacks_health-checker/thc/env.launch/health-check/THC_PATH.default
open /layers/paketo-buildpacks_health-checker/thc/env.launch/health-check/THC_PATH.default: no such file or directory
```

This impacts release-1.x as well as main.

dmikusa-pivotal:
Resolves #158 for main branch.


internal/environment_writer.go
```diff
@@ -37,6 +37,11 @@ func (w EnvironmentWriter) Write(path string, environment map[string]string) err
 
 	for key, value := range environment {
 		f := filepath.Join(path, key)
+
+		if err := os.MkdirAll(filepath.Dir(f), 0755); err != nil {
```
samj1912, line 41: Should this be outside the for loop? Isn't the directory for all the env vars the same?

dmikusa-pivotal: Normally, but when you have an environment variable that's specific to a process type, there is a subfolder. That means key here will look like this: "health-check/THC_PATH.default". It presently fails if you try to do this because the process type subfolder isn't created.

samj1912: Can we add this as a comment as well?

samj1912 (approved):

internal/environment_writer.go
dmikusa-pivotal, line 41:
```suggestion
		// required to support process-specific environment variables
		if err := os.MkdirAll(filepath.Dir(f), 0755); err != nil {
```

dmikusa-pivotal, line 41:
```suggestion
		// required to support process-specific environment variables
		if err := os.MkdirAll(filepath.Dir(f), 0755); err != nil {
```

dmikusa-pivotal: Done.

Issue #50, opened 2021-04-09
Add docs and examples for usage 


samj1912:
Currently the documentation on libcnb is a bit lacking when it comes to usage. We should add some examples on how to use the various apis/ utilities available in libcnb with a simple buildpack.

Examples of what I am expecting from the resolution of this issue -

https://github.com/paketo-buildpacks/packit#buildpack-interface

https://github.com/Malax/libcnb#example-buildpack

We could also consider adding a buildpack to the samples repository that uses libcnb.

AidanDelaney:
Please assign to me.

AidanDelaney:
We need a concrete example here.  I'm going to provide example `detect` and `build` sources for a `pip install` buildpack.  We will assume the existence of a `python-dist` buildpack that provides a runtime and a `pip-dist` buildpack that provides the `pip` binary.  In our simplified scenario `requirements.txt` can contain only package names and comment lines.  This scenario is similar to the simplified NodeJS scenario used in the packit documenteation.

Is this scenario both simple enough to understand and complex enough to demonstrate the basics of `libcnb`?

dmikusa-pivotal:
I think it would be nice to have an example that's self-contained. Having the assumed existence of other buildpacks, IMHO, makes it harder to follow & ultimately test. Unless you are going to actually use an existing buildpack for those like Paketo. That does make examples harder.

Taking a step back, thinking about the types of things I'd like to see in a demo:

- Detection
  - how to discover basic things like app location, buildpack location, etc..
  - Detection pass/fail
  - providing/requiring items in the buildplan
  - looking at bindings

- Build
  - Common locations at build (very similar to detect) & finding binding info
  - Looking at the buildplan
  - Failing a build
  - Logging/Writing messages to a user
  - Layers (creating, writing files to them, layer flags)
  - Setting labels, process types, slices and exec.d helpers

- Packaging up the buildpack
  - This is important cause you need to compile the Go code & run `pack buildpack package` against the compiled code.
  - That said, I'm not sure what to suggest. We have a system for this, which is being used with the profile buildpack but is undoubtedly more complicated than necessary. Probably, just need a basic Makefile we can suggest with some commands to test, compile & package.

I'm not sure there will be a legitimate sample buildpack that uses all that stuff, but an idea might be to branch off in a section, show one of the features like slices, say "here's how we could add a slice", but then backtrack to the main example and explain that's not necessary to add slices in this example.

[...]

PR #164, opened 2022-07-27
Provide example on using `libcnb.Detector` interface

AidanDelaney:
Show a number of examples in increasing complexity.  Each example
introduces a single idea.


AidanDelaney:
This starts to address #50.  It's not reference documentation so I don't think it fits into the https://tip.golang.org/doc/comment model.  This tutorial has a specific order to it (hence it's not reference documentation) in that it introduces very simple examples and incrementally adds complexity.

Would we prefer this tutorial-style documentation as a long comment on the [`Detector` interface](https://pkg.go.dev/github.com/buildpacks/libcnb#Detector)?  Or is this not the right approach to addressing #50?

dmikusa-pivotal:
>Would we prefer this tutorial-style documentation as a long comment on the [Detector interface](https://pkg.go.dev/github.com/buildpacks/libcnb#Detector)? Or is this not the right approach to addressing https://github.com/buildpacks/libcnb/issues/50?

I agree that it probably doesn't fit well into the Go docs. I'm sure you could stuff it into the Go docs somewhere, but there's not a clear place to put it. I'm OK with this being separate, with references to the Go docs where relevant.

I think the question to me is where to put it? The `docs/` folder in the repo is one possibility, although, I have personally never found it all that helpful when projects do this. Less because of the location, being colocated with the code is reasonable, but more because it doesn't have a cohesive flow (or good search) and often seems like an afterthought. 

If we want to keep it close to the code, I'd be more in favor of using a Github Wiki. I could also see the docs being a separate static web page, perhaps markdown generated. Any thoughts about either of those options?

AidanDelaney:
I think we should try the wiki approach.  This would allow us to document v1 and then quickly restructure to v2 when that is released.  @samj1912 do you want to enable wiki on this repo?

samj1912:
@AidanDelaney I would instead prefer if we use https://go.dev/blog/examples#larger-examples

Wikis are generally not that discoverable for go docs and also likely to get stale.

dmikusa-pivotal:
@samj1912 I like that it compiles the examples, that's nice, but how flexible is it where the examples go? 

The sample there, the sort package, just has some files referenced as examples in its package doc. 

Screen Shot 2022-07-29 at 2 08 17 PM

I'm struggling to visualize how this will fit as a doc comment. The additional context that @AidanDelaney has wrapped around the examples here is important and my concern is that you lose that with inline godoc. I'm also concerned about the volume of the content, there's quite a bit here and that's just for detect. Presumably, there will be something similar for build as well. Putting all of that into the docs is a lot of info right out of the gate. How much control do we have to organize this information?

samj1912:
I thought we agreed that libcnb repo will only contain reference examples for using various APIs and that full tutorials will end up in buildpacks.io.

I would prefer not to create yet another divergence and use Wikis for user documentation. Wikis are generally not that visible, easy to test or automate or review. 

With both the examples and the buildpacks.io website we get review and testing abilities for docs and being able to run it in a CI and they are also a lot more visible.

samj1912:
The full examples also end up in files instead of doc comments -

for eg https://cs.opensource.google/go/go/+/master:src/sort/example_keys_test.go

renders the SortKeys example above.

dmikusa-pivotal:
> I thought we agreed that libcnb repo will only contain reference examples for using various APIs and that full tutorials will end up in buildpacks.io.

I don't recall off the top of my head, mostly just reacting to the content in this PR. It seems like more tutorial and less documentation. I'm totally OK with having this on buildpacks.io docs.

AidanDelaney:
I'm adding small examples of Detect and Build here.  I'll move the tutorial style explanation to buildpacks.io/docs

AidanDelaney:
When we're happy with this approach, I'll crate 1.x versions of these examples.

AidanDelaney:
Hmm...I'm unsure why the Unit Test check is not re-running after a force push.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/buildpacks/libcnb
 go 1.17
 
 require (
+	github.com/CycloneDX/cyclonedx-go v0.6.0
```
dmikusa-pivotal, line 6: I'm reluctant to have this package added. Especially because it is just for example code. Can we do without this? Or can someone confirm that Go is smart enough to prune this out of the resulting buildpack binary image?

dmikusa-pivotal:
> Hmm...I'm unsure why the Unit Test check is not re-running after a force push.

Not 100% sure. It should re-run. Maybe try resolving the conflict with go.mod/sum?

go.mod
AidanDelaney: Go is smart enough to prune unused dependencies.  Writing a trivial `main.go` I can then run `go version` to see the dependencies.

```
$ go version -m main
main: go1.18.1
	path	command-line-arguments
	dep	github.com/BurntSushi/toml	v1.2.0	h1:Rt8g24XnyGTyglgET/PRUNlrUeu9F5L+7FilkXfZgs0=
	dep	github.com/Masterminds/semver	v1.5.0	h1:H65muMkzWKEuNDnfl9d70GUjFniHKHRbFPGBuZ3QEww=
	dep	github.com/buildpacks/libcnb	(devel)
	dep	github.com/onsi/gomega	v1.20.0	h1:8W0cWlwFkflGPLltQvLRB7ZVD5HuP6ng320w2IS245Q=
	build	-compiler=gc
	build	CGO_ENABLED=1
	build	CGO_CFLAGS=
	build	CGO_CPPFLAGS=
	build	CGO_CXXFLAGS=
	build	CGO_LDFLAGS=
	build	GOARCH=arm64
	build	GOOS=darwin
```

samj1912 (approved):

PR #227, opened 2023-04-21
Read bindings from VCAP_SERVICES

c0d1ngm0nk3y:
Buildpacks like https://github.com/paketo-buildpacks/dynatrace need specific bindings (e.g. credentials to download assets). If the bindings are not only read from the file system, but from the env variable `VCAP_SERVICES`, those buildpacks could be used not only on kubernetes, but also on cloudfoundry.

Any thoughts?

### DIFF
--- a/platform.go
+++ b/platform.go
@@ -17,6 +17,7 @@
 package libcnb
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/fs"
@@ -42,6 +43,9 @@ const (
 
 	// EnvBuildpackDirectory is the name of the environment variable that contains the path to the buildpack
 	EnvBuildpackDirectory = "CNB_BUILDPACK_DIR"
+
+	// EnvVcapServices is the name of the environment variable that contains the bindings in cloudfoundry
+	EnvVcapServices = "VCAP_SERVICES"
 
 	// EnvLayersDirectory is the name of the environment variable that contains the root path to all buildpack layers
 	EnvLayersDirectory = "CNB_LAYERS_DIR"
@@ -166,8 +170,37 @@ func NewBindingsFromPath(path string) (Bindings, error) {
 	return bindings, nil
 }
 
+type vcapServicesBinding struct {
+	Name        string            `json:"name"`
+	Credentials map[string]string `json:"credentials"`
+}
+
+// NewBindingsFromVcapServicesEnv creates a new instance from all the bindings given from the VCAP_SERVICES.
+func NewBindingsFromVcapServicesEnv(content string) (Bindings, error) {
+	var contentTyped map[string][]vcapServicesBinding
+
+	err := json.Unmarshal([]byte(content), &contentTyped)
+	if err != nil {
+		return Bindings{}, nil
+	}
+
+	bindings := Bindings{}
+	for t, bArray := range contentTyped {
+		for _, b := range bArray {
+			bindings = append(bindings, Binding{
+				Name:   b.Name,
+				Type:   t,
+				Secret: b.Credentials,
+			})
+		}
+	}
+
+	return bindings, nil
+}
+
 // NewBindings creates a new bindings from all the bindings at the path defined by $SERVICE_BINDING_ROOT.
 // If that isn't defined, bindings are read from <platform>/bindings.
+// If that isn't defined, bindings are read from $VCAP_SERVICES.
 // If that isn't defined, the specified platform path will be used
 func NewBindings(platformDir string) (Bindings, error) {
 	if path, ok := os.LookupEnv(EnvServiceBindings); ok {
@@ -176,6 +209,10 @@ func NewBindings(platformDir string) (Bindings, error) {
 
 	if path, ok := os.LookupEnv(EnvPlatformDirectory); ok {
 		return NewBindingsFromPath(filepath.Join(path, "bindings"))
+	}
+
+	if content, ok := os.LookupEnv(EnvVcapServices); ok {
+		return NewBindingsFromVcapServicesEnv(content)
 	}
 
 	return NewBindingsFromPath(filepath.Join(platformDir, "bindings"))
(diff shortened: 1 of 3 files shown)


c0d1ngm0nk3y:
@samj1912 @dmikusa This is only meant to be a rough scratch how supporting bindings from `cloudfoundry` could look like. To get some discussion started. It would enable some cnb to be used in `cloudfoundry`. WDYT?

platform.go
```diff
@@ -174,6 +206,10 @@ func NewBindings(platformDir string) (Bindings, error) {
 		return NewBindingsFromPath(path)
 	}
 
+	if content, ok := os.LookupEnv(EnvVcapServices); ok {
```
dmikusa, line 209: 1. Any objections to having this happen last? (or first). I'm not sure I like it happening in the middle of looking for service bindings. Or was there a specific reason you put this in the middle?

2. Please update the comment to reflect how things are loaded.

platform_test.go
```diff
@@ -46,6 +46,25 @@ func testPlatform(t *testing.T, context spec.G, it spec.S) {
...
+	context("Cloudfoundry VCAP_SERVICES", func() {
+		it.Before(func() {
+			content, err := os.ReadFile("testdata/vcap_services.json")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(os.Setenv(libcnb.EnvVcapServices, string(content))).To(Succeed())
+		})
+
+		it.After(func() {
+			Expect(os.Unsetenv(libcnb.EnvVcapServices)).To(Succeed())
+		})
```
dmikusa, line 58: TIP: you can use [`t.Setenv`](https://pkg.go.dev/testing#B.Setenv) and then it cleans up automatically.

testdata/vcap_services.json
```diff
@@ -0,0 +1,44 @@
...
+        "instance_name": "elephantsql-c6c60",
+        "label": "elephantsql",
+        "tags": [
+          "postgres",
+          "postgresql",
+          "relational"
+        ],
+        "plan": "turtle",
+        "credentials": {
+          "uri": "postgres://exampleuser:examplepass@babar.elephantsql.com:5432/exampleuser"
```
dmikusa, line 17:
```suggestion
          "uri": "postgres://exampleuser:examplepass@postgres.example.com:5432/exampleuser"
```

Use example.com domain for tests/documentation instead of a real domain.

testdata/vcap_services.json
```diff
@@ -0,0 +1,44 @@
...
+        "binding_name": null,
+        "instance_guid": "8c907d0f-ec0f-44e4-87cf-e23c9ba3925d",
+        "instance_name": "mysendgrid",
+        "label": "sendgrid",
+        "tags": [
+          "smtp"
+        ],
+        "plan": "free",
+        "credentials": {
+          "hostname": "smtp.sendgrid.net",
```
dmikusa, line 36:
```suggestion
          "hostname": "smtp.example.com",
```

dmikusa (changes requested):
This sounds good to me. I have a couple of minor suggestions.

I do wonder if the name/credentials will be enough. I know it was common for apps on CF to look at labels or tags to locate a service (kind of like how type is used with bindings). Maybe we should try to include that information? We could expose a `tags` and `labels` key on the binding with that information.

testdata/vcap_services.json
c0d1ngm0nk3y: I just copied the example from the cloudfoundry documentation, but that is a fair point.

c0d1ngm0nk3y:
> This sounds good to me. I have a couple of minor suggestions.

Thanks for the comments. I have uploaded another commit.

> I do wonder if the name/credentials will be enough. I know it was common for apps on CF to look at labels or tags to locate a service (kind of like how type is used with bindings). Maybe we should try to include that information? We could expose a `tags` and `labels` key on the binding with that information.

Yes, we thought about the team and wondered even what to take as `type`. The `tags` and the `label` make the usage much more flexible. But we came to the conclusion, that we do not need this flexibility for now. But adding this information to the secret to at least keep this information sounds like a good first step. I will have a look.

> Also, do you propose this for libcnb v2+ or do you intend to see it backported to v1? I think we could backport it because it's new functionality, although v1 is kind of just maintenance at this point.

I don't see a reason not to downport it, but I would probably wait if someone want/needs it there. It sounds like a fair feature to be "only" available in `v2`. WDYT?

c0d1ngm0nk3y:
> > I do wonder if the name/credentials will be enough. I know it was common for apps on CF to look at labels or tags to locate a service (kind of like how type is used with bindings). Maybe we should try to include that information? We could expose a `tags` and `labels` key on the binding with that information.

On a second thought, it might be strange to add this as long the service binding spec does not specify `tags` and `labels`. So I would propose to keep it uniform in this regard and not add this information for now. It could still easily be added later.

> > Also, do you propose this for libcnb v2+ or do you intend to see it backported to v1? I think we could backport it because it's new functionality, although v1 is kind of just maintenance at this point.
> 
> I don't see a reason not to downport it, but I would probably wait if someone want/needs it there. It sounds like a fair feature to be "only" available in `v2`. WDYT?

My bad. I re-checked and the paketo buildbacks use `v1`, so downporting it would really add some benefit/usage. What would be the correct approach?

dmikusa:
>On a second thought, it might be strange to add this as long the service binding spec does not specify tags and labels. So I would propose to keep it uniform in this regard and not add this information for now. It could still easily be added later.

I'm OK with that. If it becomes something that's needed, we can add it later.

>My bad. I re-checked and the paketo buildbacks use v1, so downporting it would really add some benefit/usage. What would be the correct approach?

Yes, at the moment Paketo is on v1. It'll probably be a little while before Paketo is on v2, so the advantage would be getting this into Paketo users' hands much quicker. It's not a big deal to backport, we'll just need a second PR against the `release-1.x` branch. It will likely be very similar, so you could try cherry-picking the commit and see if it applies cleanly. If not, then it might require a little manual work but it still shouldn't be too bad.

dmikusa (approved):

dmikusa:
This looks good to me. I'm going to give it a little bit just in case @samj1912 has any thoughts. If not, I'll merge it on Friday.

samj1912 (approved):

PR #235, opened 2023-05-30
Distinguish provider and type when reading binding from VCAP_SERVICES

c0d1ngm0nk3y:
see [this comment](https://github.com/buildpacks/libcnb/pull/228#issuecomment-1537176590).

When reading bindings from `VCAP_SERVICES`, `provider` and `type` are set and could have different values.


dmikusa:
This looks good. One question.

> I think we should also try to flatten out the credentials into key/value pairs on the binding.

Thoughts on this?

This PR has creds defined as `Credentials map[string]string`, and the entire Secret value is set to the map. This would make fetching values out of the credentials a little harder because you'd have to look for a binding of the right type, then pull the `Credentials`, then pull out the value from the map. I think this would also be problematic with auto-config Spring Cloud Bindings which expects bindings with names like `url` and `password`, not `credentials`.

I think we should keep what you have here and put it in as credentials, but we should also flatten the credentials so that there is a new (credentials key as the name and credentials value as the secret) Secret for each key in the Credentials map. That would mean you could look them up either way, fetch the whole set of credentials from one binding or fetch them individually.

My only concern was with overlap in credentials (like if something inside credentials had a key named credentials) and something outside of credentials (like type or credentials), but I don't think that should generally happen.

dmikusa:
And thanks for submitting this so quickly!

c0d1ngm0nk3y:
> I think we should keep what you have here and put it in as credentials, but we should also flatten the credentials so that there is a new (credentials key as the name and credentials value as the secret) Secret for each key in the Credentials map. That would mean you could look them up either way, fetch the whole set of credentials from one binding or fetch them individually.

That sounds good. That is what I would prefer. It "just works" in both ways.

> My only concern was with overlap in credentials (like if something inside credentials had a key named credentials) and something outside of credentials (like type or credentials), but I don't think that should generally happen.

I agree that generally should not happen, but it can of course. Maybe we can just skip any property that is in conflict (e.g. `key`). So you would have to access it via `credentials`. Maybe it is at least worth some warning since the user could face some unexpected behaviour.

c0d1ngm0nk3y:
> I think this would also be problematic with auto-config Spring Cloud Bindings which expects bindings with names like `url` and `password`, not `credentials`.

I am afraid, I didn't get it yet. Are you referring to [this](https://github.com/spring-cloud/spring-cloud-bindings)? Since `Credentials` get passed into `Secret` (see [here](https://github.com/buildpacks/libcnb/blob/d77e5b49f397146d6d27e7d62196b5f30ca4e255/platform.go#LL193C15-L193C15)), the usage should be the same.

* Find Binding
* Get Secret
* Access Property

Or what do you mean?

dmikusa:
>I am afraid, I didn't get it yet. Are you referring to [this](https://github.com/spring-cloud/spring-cloud-bindings)? 

Yes.

>Since Credentials get passed into Secret (see [here](https://github.com/buildpacks/libcnb/blob/d77e5b49f397146d6d27e7d62196b5f30ca4e255/platform.go#LL193C15-L193C15)), the usage should be the same.

Sorry. That was my misunderstanding. You're right.

dmikusa (approved):

PR #251, opened 2023-08-05
Change module versions & prep to cut an alpha

dmikusa:
This PR changes all of the module imports to use the `v2/` path.

The plan would be to merge this and then run `git tag v2.0.0-alpha.1` followed by `git push origin v2.0.0-alpha.1`.

That should tag an alpha release and make it a little easier for folks to start using the v2 branch.


dmikusa:
Does this look right @samj1912? Trying to take another step toward a v2 release.

samj1912 (approved):

Issue #254, opened 2023-09-01
Creating bindings from VCAP_SERVICES fails if credentials are nested


modulo11:
The `credentials` field contained in `VCAP_SERVICES` is [specified](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#VCAP-SERVICES) as:
> A JSON object containing the service-specific credentials needed to access the service instance

When a `Binding` is created from `VCAP_SERVICES` credentials are [expected](https://github.com/buildpacks/libcnb/blob/3e3992aaa5ccabe485b529c5ccfe6d38b7678ea6/platform.go#L179-L183) as `map[string]string` (flat) where in reality they can be nested (`map[string]any`). A `Binding` [expects](https://github.com/buildpacks/libcnb/blob/3e3992aaa5ccabe485b529c5ccfe6d38b7678ea6/platform.go#L93) them to be `map[string]string` as well. Even worse, translating `VCAP_SERVICES` into `Bindings` [swallows](https://github.com/buildpacks/libcnb/blob/3e3992aaa5ccabe485b529c5ccfe6d38b7678ea6/platform.go#L191) any marshaling error.

Simple `VCAP_SERVICES` to reproduce the issue:

```json
{
  "elephantsql-provider": [
    {
      "name": "elephantsql-binding-c6c60",
      "label": "elephantsql-type",
      "credentials": {
        "uri": "postgres://postgres.example.com",
        "foo": {
          "bar": 42
        }
      }
    }
  ]
}
```

There are a couple potential solutions that came to my mind:

- Flatten incoming credentials, which would just cut of any nested structures. From the example, only the `uri` field would be inside the `secret` field of `Binding`. This would also prevent the whole process of translating to either fully succeed or just silently fail.
- Change type of  the`secret` property in `Binding` from `map[string]string` to `map[string]any`. That would be a breaking change with lots of impact on other projects I guess.
- Treat `credentials` as raw data which is passed and "interpreted" by consuming parties. Would probably lead to even more work downstream
- Anything else?

stephprat:
Ignore nested?

stephprat:
For my use case, ignoring nested is fine.

dmikusa:
> Even worse, translating VCAP_SERVICES into Bindings [swallows](https://github.com/buildpacks/libcnb/blob/3e3992aaa5ccabe485b529c5ccfe6d38b7678ea6/platform.go#L191) any marshaling error.

Definitely a bug.

[...]

PR #256, opened 2023-09-04
Convert nested binding credentials to JSON

modulo11:
Fixes https://github.com/buildpacks/libcnb/issues/254.

Fix swallowing the parsing error and marshal properties to JSON if they are not a string.

### DIFF
--- a/platform.go
+++ b/platform.go
@@ -177,9 +177,22 @@ func NewBindingsFromPath(path string) (Bindings, error) {
 }
 
 type vcapServicesBinding struct {
-	Name        string            `json:"name"`
-	Label       string            `json:"label"`
-	Credentials map[string]string `json:"credentials"`
+	Name        string                 `json:"name"`
+	Label       string                 `json:"label"`
+	Credentials map[string]interface{} `json:"credentials"`
+}
+
+func toJSONString(input interface{}) (string, error) {
+	switch in := input.(type) {
+	case string:
+		return in, nil
+	default:
+		jsonProperty, err := json.Marshal(in)
+		if err != nil {
+			return "", err
+		}
+		return string(jsonProperty), nil
+	}
 }
 
 // NewBindingsFromVcapServicesEnv creates a new instance from all the bindings given from the VCAP_SERVICES.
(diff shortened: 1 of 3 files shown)


platform.go
```diff
@@ -177,9 +177,22 @@ func NewBindingsFromPath(path string) (Bindings, error) {
...
-	Label       string            `json:"label"`
-	Credentials map[string]string `json:"credentials"`
+	Name        string                 `json:"name"`
+	Label       string                 `json:"label"`
+	Credentials map[string]interface{} `json:"credentials"`
+}
+
+func toJSONString(input interface{}) (string, error) {
+	switch in := input.(type) {
+	case string:
```
c0d1ngm0nk3y, line 187: Is this case needed? For `string` `json.Marshal` should do fine, shouldn't it? It will handle the `int` case as well.

c0d1ngm0nk3y:
@dmikusa @samj1912 Any thoughts on this pr?

platform_test.go
```diff
@@ -55,11 +55,11 @@ func testPlatform(t *testing.T, context spec.G, it spec.S) {
...
-			Expect(bindings).To(HaveLen(2))
-			types := []string{bindings[0].Type, bindings[1].Type}
-			Expect(types).To(ContainElements("elephantsql-type", "sendgrid-type"))
-			providers := []string{bindings[0].Provider, bindings[1].Provider}
-			Expect(providers).To(ContainElements("elephantsql-provider", "sendgrid-provider"))
+			Expect(bindings).To(HaveLen(3))
+			types := []string{bindings[0].Type, bindings[1].Type, bindings[2].Type}
+			Expect(types).To(ContainElements("elephantsql-type", "sendgrid-type", "postgres"))
+			providers := []string{bindings[0].Provider, bindings[1].Provider, bindings[2].Provider}
+			Expect(providers).To(ContainElements("elephantsql-provider", "sendgrid-provider", "postgres"))
```
dmikusa, line 62:
```suggestion
			Expect(providers).To(ContainElements("elephantsql-provider", "sendgrid-provider", "postgres"))
			
			// order of bindings can change, so sorting it for consistent test results
			slices.SortFunc(bindings, func(a, b libcnb.Binding) int {
				return cmp.Compare(a.Name, b.Name)
			})

			// verify regular values
			Expect(bindings[0].Secret).To(HaveKeyWithValue("bool", true))
			Expect(bindings[0].Secret).To(HaveKeyWithValue("int", 1))
			Expect(bindings[0].Secret).To(HaveKeyWithValue("uri", "postgres://exampleuser:examplepass@postgres.example.com:5432/exampleuser"))

			// verify with nested values
			Expect(bindings[2].Secret).To(HaveKeyWithValue("username", "foo"))
			Expect(bindings[2].Secret).To(HaveKeyWithValue("password", "bar"))
			Expect(bindings[2].Secret).To(HaveKeyWithValue("urls", `{"example":"http://example.com"}`))
```

I think things are working OK, but I'd like to see the values verified by reading the properties back and confirming the correct values/types. This will just help prevent potential regressions in the code.

dmikusa:
Looks good to me, just the one suggestion with the test.

platform_test.go
pbusko: done

dmikusa (approved):

PR #266, opened 2023-10-27
Update Dependabot config

edmorley:
Dependabot updates for GitHub Actions are now enabled, for parity with the other repos.

In addition, Go minor/patch dependencies will now be grouped, using the new Dependabot grouping feature:
https://github.blog/changelog/2023-08-17-grouped-version-updates-by-semantic-version-level-for-dependabot/

Major updates, as well as security updates will still be opened as separate PRs. I've not grouped GitHub Actions update PRs, since the volume is typically much lower for those.

Lastly, the schedule has been changed from daily to weekly.

This reduces project maintenance toil (no more having to manually create combined update PRs), plus makes it less painful for contributors to subscribe to repository notifications (currently there is a lot of noise from Dependabot PRs being opened/auto-rebased etc).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,9 +3,22 @@ updates:
     - package-ecosystem: gomod
       directory: /
       schedule:
-        interval: daily
+        interval: weekly
+      groups:
+        # Group all minor/patch go dependencies into a single PR.
+        go-dependencies:
+          update-types:
+            - "minor"
+            - "patch"
       ignore:
         - dependency-name: github.com/onsi/gomega
+      labels:
+        - semver:patch
+        - type:dependency-upgrade
+    - package-ecosystem: github-actions
+      directory: /
+      schedule:
+        interval: weekly
       labels:
         - semver:patch
         - type:dependency-upgrade


dmikusa:
Presently we have this:

1. Dependabot checks daily and sends PRs
2. A workflow that checks Go modules weekly and sends a bulk update
3. All workflows are managed by `octo` (which is part of https://github.com/paketo-buildpacks/pipeline-builder/)

I agree that 1.) is unnecessary. It's rare that we have dependency updates, and most of the time, we merge in the weekly update from 2.) and that is sufficient to keep things up-to-date. I'd be fine turning this off.

The reason we don't use dependabot for weekly updates is that it's a job that we get free with `octo` pipelines, and it is a little more flexible. We can force it to run whenever, where dependabot runs on its own timeline, and also there were some dependencies that dependabot overlooks (I can't 100% remember off the top of my head, but I think it was transitive deps). We found it better to update modules directly every so often to make sure all the deps were being bumped.

The reason we don't use dependabot to keep workflows up-to-date is that `octo` does that (or rather it keeps them all in sync across all of the repos we use with `octo`, like libcnb, profile buildpack, and Paketo). If we turned this on, it would create a bot war with `octo` trying to change the versions one way and dependabot another.

edmorley:
> where dependabot runs on its own timeline,

Dependabot can be triggered manually fwiw (see https://github.com/buildpacks/lifecycle/pull/1228#issuecomment-1783103369).

The rest of the points make sense - I'll close this given that this repo seems to be managed by other tooling then.

dmikusa:
Oh, that's awesome about dependabot. Thanks for sharing that.

PR #272, opened 2024-01-18
improve SBOM validation

### LINKED ISSUE #126: Relax SBOM validation
We [validate SBOM formats in libcnb](https://github.com/buildpacks/libcnb/blob/main/build.go#L393), but this only takes into consideration the core mime type, not any additional parameters/specifiers.

We need to improve the validation to take into consideration additional parameters or relax it so that they are ignored.

sagnik3788:
Resolves #126


dmikusa:
Can you add a test case [here](https://github.com/buildpacks/libcnb/blob/main/build_test.go#L643) that proves this change fixes the issue?

The idea would be to do something like change `sbom-formats = ["application/vnd.cyclonedx+json"]` to `sbom-formats = ["application/vnd.cyclonedx+json;charset=us-ascii"]`. I don't know if that's valid, but I'm pretty sure that will fail using the current methods. Because it will detect the media type as one of these options (https://github.com/buildpacks/libcnb/blob/main/layer.go#L96-L101), which don't match exactly.

dmikusa:
Hi @sagnik3788, just wanted to follow up and see how things are going. Did you see my previous note? Are you still interested in submitting a code change here? Thanks!

dmikusa:
Closing as we need some test coverage to be able to merge this. @sagnik3788 if you want to pick this back up feel free to send a new PR. Thanks

PR #284, opened 2024-04-19
Do not enforce presence of CNB_STACK_ID

c0d1ngm0nk3y:
If`CNB_STACK_ID` is not present, the buildpacks should not just fail.

related rfc: https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md
first step for fixing https://github.com/paketo-buildpacks/libpak/issues/329

This also aligns the behavior with `packit`, so that all `paketo` buildpacks behave similar.

Downport of https://github.com/buildpacks/libcnb/pull/283


dmikusa:
All set. Thanks for this. Let me know when you want me to cut a release with this. I can do that anytime.

Issue #260, opened 2023-10-03
Add support for buildpacks API 0.10


dmikusa:
This could probably be broken down into more tasks, but at a high level we need to add support for 0.10.

https://buildpacks.io/docs/reference/spec/migration/buildpack-api-0.9-0.10/

loewenstein:
Doesn't it need to be about 0.8 -> 0.9 -> 0.10, i.e. using https://buildpacks.io/docs/reference/spec/migration/buildpack-api-0.8-0.9/ and then https://buildpacks.io/docs/reference/spec/migration/buildpack-api-0.9-0.10/?

nicolasbender:
> This could probably be broken down into more tasks

Could you provide more details for these subtasks, please? We haven't found any relevant changes from 0.9 -> 0.10 affecting libcnb.
Out of interest: Why was `v2` of `libcnb` created in the first place?

dmikusa:
@loewenstein 0.9 is already supported, so we just need to add 0.10 support. It needs to be an additive change.

@nicolasbender 

- [ ] We need to mark anything related to stacks as deprecated. I haven't looked to see exactly what's in there, but there is [some stuff](https://github.com/search?q=repo%3Abuildpacks%2Flibcnb%20stack&type=code).
- [ ] We need to add support for targets. This requires [toml, struct, and serialization/deserialization changes](https://buildpacks.io/docs/for-buildpack-authors/how-to/migrate/buildpack-api-0.9-0.10/#before-build). 
- [ ] There are new target-related env variables exposed. With [stacks we put that into the context](https://github.com/buildpacks/libcnb/blob/main/build.go#L228-L232), we should probably do something similar with targets.
- [ ] The migration guide doesn't [elaborate on runtime image extensions](https://buildpacks.io/docs/for-buildpack-authors/how-to/migrate/buildpack-api-0.9-0.10/#run-image-extensions-is-supported-experimental), but this needs to be spiked to see if there are any required changes to support this. For example, there might be struct/toml changes.

> Out of interest: Why was v2 of libcnb created in the first place?

In a nutshell, breaking API changes. We reworked the core build/detect mechanism. It was previously done with structs at its heart, and we changed it to be function-based now. I believe it'll be more flexible, but this was also done in conjunction with the Paketo team to position them to be able to use libcnb as the base for their packit library.

Aside from that, we've removed some other cruft like old buildpack API compatibility.

loewenstein:
[...]

PR #291, opened 2024-05-15
Support Buildpack API 0.10

c0d1ngm0nk3y:
see [task list](https://github.com/buildpacks/libcnb/issues/260#issuecomment-2110572119)

Fixes https://github.com/buildpacks/libcnb/issues/260

### DIFF
--- a/build.go
+++ b/build.go
@@ -55,7 +55,7 @@ type BuildContext struct {
 	// Platform is the contents of the platform.
 	Platform Platform
 
-	// StackID is the ID of the stack.
+	// Deprecated: StackID is the ID of the stack.
 	StackID string
 }
 
--- a/build.go
+++ b/build.go
@@ -87,7 +87,7 @@ const (
 	MinSupportedBPVersion = "0.8"
 
 	// MaxSupportedBPVersion indicates the maximum supported version of the Buildpacks API
-	MaxSupportedBPVersion = "0.9"
+	MaxSupportedBPVersion = "0.10"
 )
 
 // NewBuildResult creates a new BuildResult instance, initializing empty fields.
--- a/build.go
+++ b/build.go
@@ -57,6 +57,12 @@ type BuildContext struct {
 
 	// Deprecated: StackID is the ID of the stack.
 	StackID string
+
+	// TargetInfo contains info of the target (os, arch, ...).
+	TargetInfo TargetInfo
+
+	// TargetDistro is the target distribution (name, version).
+	TargetDistro TargetDistro
 }
 
 // BuildResult contains the results of detection.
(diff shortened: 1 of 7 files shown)


loewenstein-sap:
@samj1912 & @dmikusa could you have a look. Should be the last must have for cutting `libcnb/v2` and getting rid of branch maintenance.

loewenstein-sap:
@samj1912 & @dmikusa ^^

dmikusa:
What's here looks good. I'm still picking through the spec changes, especially the extension images part to make sure we got everything.

One thing I don't see is support for [extension config](https://github.com/buildpacks/spec/blob/main/image_extension.md#extend-configtoml-toml). This should have a struct so we can serialize/deserialize it. Probably also a new API in generate.go that allows one to easily add this configuration.

I also feel like we should be doing something with the dockerfile generation. The `GenerateResult` could have fields for this, perhaps each one could be a io.Writer. Then after the generate function runs, libcnb could take these writers and generate the files in the correct place. I feel like we should be doing something to help enforce some of the requirements of the spec in terms of what you can/can't do in the dockerfiles, however, we can probably leave that out for now. This is still experimental so our support doesn't need to be 100%.

Maybe something similar with the context folders too? https://github.com/buildpacks/spec/blob/main/image_extension.md#context-folders Although, I'm still trying to understand the purpose of those and how they might be used.

pbusko:
@dmikusa support for dockerfiles and extend-config has been added

dmikusa:
Thanks @pbusko! I'm going to do a little testing with this and I'll get back to you soon. In the meantime, when you have a second, can you bump the PR one more time. Thanks

dmikusa (approved):

dmikusa:
@pbusko - FYI, I cut https://github.com/buildpacks/libcnb/releases/tag/v2.0.0-alpha.3 which has this merged.

dmikusa:
> @dmikusa is there anything else blocking the `v2.0.0` release?

@pbusko I'm doing some cleanup on issues, and some code review to investigate that. I think we're in pretty good shape though.

Off the top of my head, the one area I would like to put a little more effort is to add some convenience around the Dockerfile/extension abstraction. The byte[] we have now is literally the least we can do. I think we can add some more convenience on top of that without too much fuss, maybe a Writer or something to facilitate higher-level interactions.

loewenstein-sap:
> Off the top of my head, the one area I would like to put a little more effort is to add some convenience around the Dockerfile/extension abstraction. The byte[] we have now is literally the least we can do. I think we can add some more convenience on top of that without too much fuss, maybe a Writer or something to facilitate higher-level interactions.

Could convenience not easily go into `2.1` instead of holding back `v2`?

dmikusa:
Sorry, but I'm going to answer your question with a question. Why rush?

loewenstein-sap:
Looking at how long the branch was open and the hassle a long running branch brings to the table I thought getting rid of it should ideally not be deferred unnecessarily.

PR #329, opened 2025-06-26
Parse type field from VCAP_SERVICES credentials

tomkennedy513:
- currently the VCAP_SERVICES parsing will use the label from the binding as the type. This is problematic because user-provided bindings will always have label 'user-proved' which prevents users from creating custom bindings that are of a known type


dmikusa:
As an FYI, v1 not being maintained anymore. The only reason I would cut another release is if there is a critical security issue that impacts the library.

PR #334, opened 2025-09-17
Enable VCAP_SERVICES_FILE_PATH file based bindings support

kiril-keranov:
The change is with regard to the [VCAP_SERVICES_FILE_PATH](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#VCAP-SERVICES-FILE-PATH) functionality introduced in CF in order to be able to store **VCAP_SERVICES** content in a file. 
The 130KB limitation for **VCAP_SERVICES** is insufficient for increasing amount of use cases, especially when having huge certificates content. This limit is increased to 1MB for **VCAP_SERVICES_FILE_PATH** content.

### DIFF
--- a/platform.go
+++ b/platform.go
@@ -49,6 +49,9 @@ const (
 
 	// EnvVcapServices is the name of the environment variable that contains the bindings in cloudfoundry
 	EnvVcapServices = "VCAP_SERVICES"
+	
+	// EnvVcapServicesFilePath is the name of the environment variable that contains the filepath bindings in cloudfoundry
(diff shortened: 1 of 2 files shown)


modulo11:
Looking at the [code](https://github.com/cloudfoundry/cloud_controller_ng/blob/3a98e06b125b1103636296a9eaf86f675baef902/app/presenters/system_environment/system_env_presenter.rb#L11-L16) in the cloud controller, there can only be one of `SERVICE_BINDING_ROOT`, `VCAP_SERVICES_FILE_PATH` or `VCAP_SERVICES`.

dmikusa (approved):

